PR #2, opened 2020-12-08
Evaluation of the Filecoin paper

markspanbroek:


evaluations/filecoin.md
```diff
@@ -0,0 +1,48 @@
...
++ Micropayments via payment channels (§5.3.1).
++ Integration with other blockchain systems such as Ethereum (§7.2) are being
+  worked on.
+
+#### Cons:
+
+- Filecoin requires its own very specific blockchain, which influences a lot
+  of its design. There is tight coupling between the blockchain, storage
+  accounting, proofs and markets.
+- Proof of spacetime is much more complex than simple challenges, and only
```
dryajov, line 29: This is a focal point of Filecoin, which emphasizes physical storage proofs over a more probabilistic approach to data availability (random sampling + erasure coding or other error correcting techniques). 

On the surface it's a sound proposition that should shield the user from data loss due to scammy providers that don't possess the actual physical storage, but does this actually mitigate the underlying issue of data loss or is it a red herring that adds little security while incurring in a large implementational and operational overhead?

PR #5, opened 2021-01-06
Evaluation of the IPFS paper

markspanbroek:


evaluations/ipfs.md
```diff
@@ -0,0 +1,41 @@
...
++ Uses proof-of-work in S/Kademlia to discourage Sybil attacks (§2.1.3)
++ Favors self-describing values such as multihash (§3.1) and multiaddr (§3.2.1)
++ BitSwap protocol for exchanging blocks supports multiple strategies (§3.4.2)
++ Uses content addressing (§3.5)
++ The Merkle DAG is simple, yet allows constucting filesystems,
+  key-value stores, databases, messaging system, etc.. (§3.5)
+
+#### Cons:
+
+- Kademlia prefers long-lived nodes, which is not ideal for mobile enviroments
```
dryajov, line 27: This is a consequence of using libp2p, which doesn't support "packet switching", it uses streaming. This can and will be alleviated with the use of quic.

evaluations/ipfs.md
```diff
@@ -0,0 +1,41 @@
...
+  (§2.1.1)
+- BitSwap falls just short of introducing a currency with micro payments,
```
dryajov, line 29: bitswap is a pluggable mechanism that in it's default implementation doesn't have any notion of payments, but can be extended to support one.

evaluations/ipfs.md
```diff
@@ -0,0 +1,41 @@
...
+  necessitating additional work for nodes to find blocks to barter with (§3.4)
+- Merkle DAGs and Paths (§3.5.1) feel like an unnecessary complication at the
```
dryajov, line 31: The merkle dag serves as a way of communicating metadata about the chuks, but it's flexible enough to support more sophisticated things. So the merkle dag doesn't actually force anything on the implementation, but it is trivial to implement a virtual unix like file system on top of it for example.

evaluations/ipfs.md
```diff
@@ -0,0 +1,41 @@
...
+  protocol level; they could have been implemented on top of a system consisting
+  solely of content addressable chunks
+- Object pinning (§3.5.3) inevitably leads to centralized gateways to IPFS, such
```
dryajov, line 34: this is a controversial "issue" that has been brought up a few times, but in reality there aren't any technical limitations that prevent removing or modifying this behavior.

evaluations/ipfs.md
```diff
@@ -0,0 +1,41 @@
...
+  as Infura and Pinata
+- There are no self-describing multiformats for encryption and signing (§3.5.5)
```
dryajov, line 36: there are and are communicated as part of the [CID](https://docs.ipfs.io/concepts/content-addressing/#identifier-formats) using [multicodecs](https://github.com/multiformats/multicodec).

evaluations/ipfs.md
```diff
@@ -0,0 +1,41 @@
...
+- IPFS uses variable size blocks instead of fixed-size chunks (§3.6)
```
dryajov, line 37: I don't think this is a hard impositions, CID can be anything and lengths can be anything as well.

evaluations/ipfs.md
```diff
@@ -0,0 +1,41 @@
...
+- Supporting version control directly in IPFS feels like an unnecessary
```
dryajov, line 38: Not sure what you mean by this, any content address store implicitly supports some level of version control, because event slightly different content will produce completely different identifiers?

evaluations/ipfs.md
```diff
@@ -0,0 +1,41 @@
...
+  complication (§3.6.5)
+- Handles mutable state by hacking it into the routing DHT (§3.7.1), instead of
```
dryajov, line 40: I don't think this is accurate, IPFS does precisely this, it uses something called [IPNS](https://docs.ipfs.io/concepts/ipns/).

evaluations/ipfs.md
markspanbroek: Quic will probably help, but you'd still need to update the DHT when you change ip address.

evaluations/ipfs.md
markspanbroek: Yes, it's indeed great that it has multiple strategies, I added that to the 'Pros' list.

evaluations/ipfs.md
markspanbroek: Yes, the Merkle DAG is very versatile, and I like that. The point that I was trying to make is that you could probably leave the Merkle DAG out of the base protocol, thereby simplifying it. And then build a layer on top that implements the Merkle DAG.

evaluations/ipfs.md
markspanbroek: Yes, I'd be in favor of removing it.

evaluations/ipfs.md
dryajov: I don't think there is an efficient routing and discovery mechanism besides a DHT, so it's a bit of a necessary tradeoff. Also, you don't necessarily update the DHT, once you've changed networks all you're previous connections have died and you need to do another round of discovery at which point other nodes learn your new IP.

evaluations/ipfs.md
dryajov: The merkle DAG is mentioned in the paper, but it is indeed a separate component which can be swapped by anything else. The point on bringing it up in the paper (I presume), is to give a full view of an efficient assembly of this components.

evaluations/ipfs.md
dryajov: or, you broadcast your new address to the nodes already present in your buckets.

evaluations/ipfs.md
markspanbroek: I looked at those, but couldn't find any encryption or signature schemes in the [multicodecs table](https://github.com/multiformats/multicodec/blob/master/table.csv). I also saw a bunch of open issues in the multiformat repo that lead me to conclude that it's not there yet.

evaluations/ipfs.md
markspanbroek: No indeed, none of these are hard impositions. Just that fixed-size chunks might make a marketplace and incentives mechanism a bit simpler.

evaluations/ipfs.md
markspanbroek: In paragraph 3.6 they describe a model that is very close to Git, including things like commits. I don't think that this is a necessary component of a decentralized storage protocol, but could better be built on top of it.

evaluations/ipfs.md
markspanbroek: In the paper they describe IPNS as follows: "Because  this  is not a  content-addressed  object,  publishing it relies on the only mutable state distribution system in IPFS, the Routing system.  The process is (1)  publish  the  object  as  a  regular  immutable  IPFSobject, (2) publish its hash on the Routing system as a metadata value"

evaluations/ipfs.md
markspanbroek: Yes, indeed. Kademlia is probably the DHT to go with, however there are [other DHTs](https://aircconline.com/ijcnc/V9N5/9517cnc04.pdf) that claim to handle churn better.

evaluations/ipfs.md
markspanbroek: In paragraph 3.5 they mention that every `IPFSObject` has an array of links (forming the DAG), from which I concluded that it's part of the core of IPFS.

evaluations/ipfs.md
dryajov: Yes, IPNS is essentially the peer's public key that can be associated with any other content id, this public keys serves as an immutable namespace for the content associated with that peer, the namespace can be updated to point to new/different content id hashes.

The namespace can also be associated with any name resolution system, such as DNS or ENS for example. IMO, it's a much more flexible approach.

evaluations/ipfs.md
dryajov: Yes, thats misleading as there is no hard requirement for it to be so. But at its core, IPFS is a lego of components (similar to libp2p), that allows swapping different subsystems for other. That said, I like the way the Merkle DAG interacts with the rest of the system.

evaluations/ipfs.md
markspanbroek: I would also be in favor of using DNS or ENS to directly point to an immutable IPFS object. If you then need to update some content, you simply update DNS/ENS to point to a new Merkle root.

dryajov: This breaks for frequently changing content, for example if you have a feed that gets update frequently you need to update the DNS or ENS every time as well, whereas with IPNS, you simply "push" the new hash.

markspanbroek: Good point, I hadn't considered that.

evaluations/ipfs.md
dryajov: `multihash` describes different cryptographic functions, not sure what's missing there? 

>  I also saw a bunch of open issues in the multiformat repo that lead me to conclude that it's not there yet.

Multiformats are used pretty extensively across the ecosystem. For example, libp2p implementation uses multiaddrs exclusively. It is pretty stable. I haven't looked, but a lot of the issues might be related to adding support for new formats?

markspanbroek: I'd expect to be able to specify that an IPFS object is encrypted with Salsa20 and authenticated with Poly1305 (for example). Having upgradable encryption schemes feels like a natural fit for multiformats, yet I could not find a multiformat for encryption. Multihash for instance only only describes cryptographic hashing functions, not encryption schemes.

dryajov: That would be [multicodec](https://github.com/multiformats/multicodec), it describes the contents "type", with the multicodec [table](https://github.com/multiformats/multicodec/blob/master/table.csv), defining all the supported codecs.

If you read the table it mostly describes application specific codecs, for example ethereum datastructures - https://github.com/multiformats/multicodec/blob/master/table.csv#L54-L62. 

It is possible to store encrypted data of course, but, it would be application specific. This is because there are many ways of doing encryption and hashing/digesting, I suspect that catering to each and every use is a bit too much. For example, where and how to locate the public key/keys, is it included in the blob, or is it stored separately? Also, if the content is encrypted, do you really want to reveal how it was encrypted?

dryajov: Interestingly, draft 3 of the paper describes a native encrypted DAG node type (not sure if it's present in earlier versions). I'll add it here since it's small enough:


--------------------------

### 3.5.5 Object-level Cryptography

IPFS is equipped to handle object-level cryptographic operations. An encrypted or signed object is wrapped in a
special frame that allows encryption or verification of the raw bytes.

```go
type EncryptedObject struct {
  Object []bytes
  // raw object data encrypted
  Tag []bytes
  // optional tag for encryption groups
}

type SignedObject struct {
  Object []bytes
  // raw object data signed
  Signature []bytes
  // hmac signature
  PublicKey []multihash
  // multihash identifying key
}
```

Cryptographic operations change the object’s hash, defining a different object. IPFS automatically verifies signatures, and can decrypt data with user-specified keychains. Links of encrypted objects are protected as well, making traversal impossible without a decryption key. It is possible to have a parent object encrypted under one key, and
a child under another or not at all. This secures links to shared objects.

evaluations/ipfs.md
dryajov: I agree, this is overkill.

evaluations/ipfs.md
markspanbroek: You're right, multicodec is probably the correct way to specify which encryption scheme is used for encrypted content. 

Here's a link to some discussions about signatures and keys in the multiformats repo: [multikey](https://github.com/ipfs/specs/issues/58), [multisig](https://github.com/multiformats/multiformats/issues/23).

markspanbroek:
@dryajov, I updated the evaluation based on your comments, what do you think?

evaluations/ipfs.md
```diff
@@ -0,0 +1,41 @@
...
+#### Cons:
+
+- Kademlia prefers long-lived nodes (§2.1.1), which is not ideal for mobile
+  enviroments (although it's unclear whether there are any better alternatives)
+- The default BitSwap strategy falls just short of introducing a currency with
+  micro payments, necessitating additional work for nodes to find blocks to 
+  barter with (§3.4)
+- Object pinning (§3.5.3) inevitably leads to centralized gateways to IPFS, such
+  as Infura and Pinata
+- There are no self-describing multiformats for encryption and signing (§3.5.5),
```
dryajov, line 35: There is, it is described in `3.5.5 Object-level Cryptography`, for both signed and encrypted data.

markspanbroek: In 3.5.5 they describe a format for encryption and signatures, but I would not consider them to be self-describing. You can’t figure out which encryption/signature algorithm to use just by looking at an `EncryptedObject` or `SignedObject`, which is what I’d expect from a self-describing format.

dryajov: Well, `SignedObject` certainly does include all the info required to identify the scheme and verify it with the public key pointed to by the `PublicKey []multihash` field. For `EncryptedObject` you are right that it's impossible to tell what crypto primitive was used, but it seems to be a deliberate omission, you probably don't want to leak too much info about the encryption scheme.

```go
type SignedObject struct {
  Object []bytes
  // raw object data signed
  Signature []bytes
  // hmac signature
  PublicKey []multihash
  // multihash identifying key
}
```

markspanbroek: I still don't quite understand how the multihash for the public key tells you the signature scheme, but fair enough, there is much more info there than I initially thought. I'll remove the point, because it's now more of a nit-picking comment than an actual downside of using IPFS.

PR #14, opened 2021-06-21
A state channel committee scheme

markspanbroek:
Contains a scheme for state channel disputes that improves on the state of the art PISA and BRICK protocols.

Early draft. Comments are very welcome.

### DIFF
--- a/design/state-channel-committee.md
+++ b/design/state-channel-committee.md
@@ -1,0 +1,122 @@
+A State Channel Committee Scheme
+================================
+
+It should in general be possible to close payment and state channels when one of
+its participants goes offline. This presents a problem: how do we know that the
+single party that attempts to close the channel presents the *latest* state of
+the channel. It could be beneficial to present an older state, where more money
+was allocated to that party.
+
+Most channel schemes therefore introduce a dispute period; a period of time in
+which the state at which a channel is closed can be disputed. During the dispute
+period the other party can prove that a state is old by providing a newer state
+that was signed by all parties.
+
+But dispute periods do not help when they are needed the most; when a party goes
+offline. An offline party cannot participate in the dispute mechanism.
+
+Schemes such as [PISA][1] solve this problem by introducing a monitoring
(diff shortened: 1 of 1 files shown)


design/state-channel-committee.md
```diff
@@ -0,0 +1,122 @@
...
+We improve on the PISA and BRICK protocols by removing the need for extra
+unidirectional payment channels to pay the monitoring service / committee.
+Because we require all state updates to be co-signed by the committee they can
+ensure that fees are paid there.
+
+We improve on the BRICK protocol by allowing signatures from all participants
+and committee members to be combined into a single signature using Schnorr or
+BLS signatures.
+
+We improve on the BRICK protocol by requiring less interaction with the
```
dryajov, line 101: - Who would be the submitter for the transaction?
- How/who pays the fees of the on chain transaction?

One possibility might be to randomly select a submitter from the committee and subsequently, split the fees of the transaction evenly across all the participants. 

I've ran into a similar issue when thinking about how to aggregate and submit proofs of storage to the chain and seems to be recurrent in all aggregation based schemes.

markspanbroek: Those are interesting questions, thanks!

In this scheme it is important that anyone can step out at any moment. Participants should be able to get out when the other party goes offline, and committee members should be able to get out if they want to retrieve their collateral (after a while). Which means that we can't select a specific or random party to submit the transaction.

But it would be unfair to have the costs of the transaction fall to a single party, so perhaps it would be best to pre-pay the closing transaction fee when opening the channel. The participants in the channel (but not the committee) would each pay their share of the closing transaction fee to the contract when they open the channel. Whoever submits the closing transaction can then be reimbursed for the gas costs by the contract.

One complication could be the fluctuating gas costs; the actual costs of closing are not known in advance. Maybe the pre-paid gas costs could be calculated pessimistically, and any funds that remain after paying for gas could be returned to the participants.

PR #15, opened 2021-08-17
Filecoin update

Menduist:
AFAICT, 
```- Proof of spacetime is much more complex than simple challenges, and only
  required to make the blockchain work (§3.3, §6.2)```
If this is implying that the PoST is only used for the consensus mechanism of the blockchain, it's not true:
Every block (~30s), a proof of storage is generated locally, and then every 24h they're all merged into a PoST and committed the the blockchain.
So the PoST is not only used to make the blockchain work, it's more importantly used to publish the proofs in a compact format.

Though I do wonder, what stop the users of retroactively computing the PoST (inside the 24h window)? I could shut down my hard drive for 23h, spin it up, compute all the proofs of the last 24h and send it to the blockchain?
Maybe it's explained somewhere and I missed it.

Last paragraph source (not in the whitepaper): https://docs.filecoin.io/about-filecoin/faq/#even-though-filecoin-will-be-backed-up-to-our-normal-ipfs-pinning-layer-we-still-need-to-know-how-quickly-we-can-access-data-from-the-filecoin-network-how-fast-will-retrieval-be-from-the-filecoin-network

### DIFF
--- a/evaluations/filecoin.md
+++ b/evaluations/filecoin.md
@@ -26,8 +26,6 @@ Dagger. It is not meant to be a criticism of Filecoin.
 - Filecoin requires its own very specific blockchain, which influences a lot
   of its design. There is tight coupling between the blockchain, storage
   accounting, proofs and markets.
-- Proof of spacetime is much more complex than simple challenges, and only
-  required to make the blockchain work (§3.3, §6.2)
 - A miners influence is proportional to the amount of storage they're
   providing (§1.2), which is an incentive to become big. This could lead to
   the same centralization issues that plague Bitcoin.
@@ -46,3 +44,6 @@ Dagger. It is not meant to be a criticism of Filecoin.
   attacks.
 - The addition of smart contracts (§7.1) feels like an unnecessary
   complication.
+- The data are stored in a non-readable (sealed) format (§3.4.2).
+  If the hosts doesn't store data in a duplicated, plain format,
+  the unsealing required to read them take a long time (1-5 hours).

### CODE REVIEW

**Comment by Menduist** [comment]
It may also be worth mentioning that:
```
+ The network actively tries to repair storage faults by introducing new
  orders in the storage market. (§4.3.4).
```
Is very vague in the whitepaper, and no implementation has currently been done.
The plan for integration is very different and more complex that what is explained in the whitepaper.

But since this is supposed to be an evaluation of the whitepaper, not sure if it belongs here.

**Comment by dryajov** [comment]
> Though I do wonder, what stop the users of retroactively computing the PoST (inside the 24h window)? I could shut down my hard drive for 23h, spin it up, compute all the proofs of the last 24h and send it to the blockchain?


PoST are computationally expensive and recursive (VDF?), you need the proof from the previous run to generate the next one. We intend to use something similar for our proofs as well. 

As far submitting the proofs, I have the same intuition, even tho you can't generate the proofs at once, because that would be computationally infeasible, revealing the proofs once every 24h limits data loss detection to the same time frame (i.e. 24h window), which limits reaction time and recoverability?

File: evaluations/filecoin.md [hunk 1]
```diff
@@ -26,8 +26,6 @@ Dagger. It is not meant to be a criticism of Filecoin.
 - Filecoin requires its own very specific blockchain, which influences a lot
   of its design. There is tight coupling between the blockchain, storage
   accounting, proofs and markets.
-- Proof of spacetime is much more complex than simple challenges, and only
```
dryajov, line 29: I think a better wording would be that "proofs of space time are (by design) computationally intensive and non-parallelizable (afaik)"

markspanbroek:
> If this is implying that the PoST is only used for the consensus mechanism of the blockchain, it's not true

You're correct, it does serve a purpose. What is hinted at here, if I remember correctly, is that the proofs in themselves aren't very useful if you can't retrieve the data. Because this is already addressed in on of the next points about incentives, it's good to change the wording here.

PR #66, opened 2022-03-01
A design for timing of storage proofs

markspanbroek:
Added a document that describes the current design of the proof timing in [our smart contracts](https://github.com/status-im/dagger-contracts), the problems that we encounter, and a new design that intends to solve these problems.

### DIFF
--- a/design/storage-proof-timing.md
+++ b/design/storage-proof-timing.md
@@ -1,0 +1,173 @@
+Timing of Storage Proofs
+========================
+
+We present a design that allows a smart contract to determine when proofs of
+storage should be provided by hosts.
+
+Context
+-------
+
+Hosts that are compensated for providing storage of data, are held accountable
+by providing proofs of storage periodically. It's important that a host is not
+able to pre-compute those proofs, otherwise it could simply delete the data and
+only store the proofs.
+
+A smart contract should be able to check whether those proofs were delivered in
+a correct and timely manner. Either the smart contract will be used to perform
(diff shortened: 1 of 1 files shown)


design/storage-proof-timing.md
```diff
@@ -0,0 +1,173 @@
...
+     m           255                      ^           0
+     e                                    |
+     |                                  pointer
+     |
+     |      - - - |-----------------------------------|
+     |           255                  ^               0
+     |                                |
+     v                             pointer
+
+Pointer duos
```
dryajov, line 123: What if we simply use a smaller modulo, wouldn't that give us a window that will keep the pointer valid for that many blocks - `WINDOW = 256 - (blocknumber + period) % modulo`? For example, if the modulo is 128, then this gives the node a `WINDOW` of at least 128 blocks where the pointer is still valid. Once the `WINDOW` has expired the proof is impossible to verify? We can adjust the `WINDOW` depending on the average block production cadence or some other parameter, even wall clock?

dryajov: Obviously we adjust the window by tweaking the `modulo`.

markspanbroek:
Replaced the pointer duos by a strategy in which the pointer is ignored some of the time, as suggested by @cskiraly.

Menduist:
If you ever feel bad about the wasted time, remember the VGA blanking time:
https://i.stack.imgur.com/6wpSA.jpg

PR #83, opened 2022-05-26
Proposal for a new design for marketplace

markspanbroek:
Draft of a new design for a marketplace that is simpler, and includes incentives for repair.

### DIFF
--- a/design/marketplace.md
+++ b/design/marketplace.md
@@ -78,8 +78,8 @@ Initially all host slots are empty. An empty slot can be filled by anyone by
 submitting a correct storage proof together with collateral.
 
 
-        collateral                              collateral
-          proof    proof          missed          proof                missed
+        proof &                                 proof &
+      collateral   proof          missed      collateral               missed
             |        |              |               |                    |
             v        v              v               v                    v
             -------------------------------------------------------------------
@@ -88,7 +88,7 @@ submitting a correct storage proof together with collateral.
                                     |                                    |
                                     v                                    v
                                 collateral                           collateral
-                                  burned                               burned
+                                  lost                                 lost
 
 
 
--- a/design/marketplace.md
+++ b/design/marketplace.md
@@ -247,9 +247,9 @@ A host is allowed to fill a slot when the distance between its id and the start
 point is less that the allowed distance.
 
                                  start point
-                                      |
+                                      |           Kademlia distance
                 t=3    t=2    t=1     v
-    -------------(------(------(------·------)------)------)--------------
+          <------(------(------(------·------)------)------)------>
                           ^                            ^
                           |                            |
                      this host is                 this host is
--- a/design/marketplace.md
+++ b/design/marketplace.md
@@ -63,6 +63,7 @@ including a small amount of host losses that can be tolerated.
       proof parameters                # proof of retrievability parameters
       erasure coding                  # erasure coding parameters
       dispersal                       # dispersal parameter
+      repair reward                   # amount of tokens paid for repairs
 
       hosts                           # amount of storage hosts (including loss)
       loss                            # number of allowed host losses
@@ -100,11 +101,11 @@ for every second of the interval a certain amount of tokens are awarded to the
 host. Hosts that fill a slot are required to submit frequent proofs of storage.
 
 When a certain number of proofs is missed, the slot is considered empty again.
-The collateral associated with the slot is used to pay a fee to the node that
-indicated that proofs were missing. The rest of the slot collateral is reserved
-for repairs. An empty slot can be filled again once another host submits a
-correct proof together with collateral. Payouts for the time interval that a
-slot is empty are burned.
+The collateral associated with the slot is mostly burned. Some of it is used to
+pay a fee to the node that indicated that proofs were missing, and some of it is
+reserved for repairs. An empty slot can be filled again once another host
+submits a correct proof together with collateral. Payouts for the time interval
+that a slot is empty are burned.
 
 Payouts for all hosts are accumulated in the contract and payed out at contract
 end. This is to ensure that the incentive posed by the collateral is not
@@ -164,18 +165,18 @@ The client is able to retrieve any funds that are left in the contract.
 Repairs
 -------
 
-When a slot becomes empty, the remaining collateral associated with the slot is
-used as an incentive to repair the lost content. Repair typically involves
-downloading other parts of the content and using erasure coding to restore the
-missing parts. This incurs costs for a host. To compensate the host for these
-costs it receives not only its own collateral back at the end of the contract,
-but also the remaining collateral from the host that failed to hold a slot.
+When a slot becomes empty, some of the remaining collateral associated with the
+slot is used as an incentive to repair the lost content. Repair typically
+involves downloading other parts of the content and using erasure coding to
+restore the missing parts. This incurs costs for a host. To compensate the host
+for these costs it receives not only its own collateral back at the end of the
+contract, but also a reward. The size of the reward is a parameter of the
+contract.
 
-We expect the collateral to be significantly higher than the costs of repair.
-This means that hosts in the network can benefit greatly from repairs, and they
-may prioritize repairs over filling slots in new contracts. This is intentional,
-we want the network to prioritize honoring existing contracts over starting new
-ones.
+The size of the reward should be chosen carefully. It should not be too low, to
+incentivize hosts in the network to prioritize repairs over filling new slots in
+the network. It should also not be too high, to prevent malicious nodes in the
+network to try to disable hosts in an attempt to collect the reward.
 
 Renewal
 -------
--- a/design/marketplace.md
+++ b/design/marketplace.md
@@ -98,21 +98,11 @@ The time interval that a slot is filled by a host determines the host payout;
 for every second of the interval a certain amount of tokens are awarded to the
 host. Hosts that fill a slot are required to submit frequent proofs of storage.
 
-When a proof is missed, the collateral associated with a slot is mostly burned,
-except for a fee for the one who marked the proof as missing. The slot is now
-considered empty again until another host submits a correct proof together with
-collateral.
-
-To incentivize repair, we allow the host that takes over a slot to count the
-interval between the missed proof and the new proof towards its own profits.
-This means that when a proof is missed, the incentive to repair increases over
-time.
-
-The situation where the payout on repair becomes larger than the required
-collateral is an interesting one. On the one hand, this could be an opportunity
-for zero-cash entry into the network. On the other hand, it might lead to
-undesired behavior because the loss of collateral is no longer that important
-for a node. This is probably worth looking further into.
+When a proof is missed, the collateral associated with a slot is used to pay a
+fee to the one who marked the proof as missing. The rest of the slot collateral
+is reserved for repairs. The slot is now considered empty again until another
+host submits a correct proof together with collateral. Payouts for the time
+interval that a slot is empty are burned.
 
 Contract lifecycle
 ------------------
@@ -164,6 +154,22 @@ The client is able to retrieve any funds that are left in the contract.
     | ended |
     ---------
 
+
+Repairs
+-------
+
+When a slot becomes empty, the remaining collateral associated with the slot is
+used as an incentive to repair the lost content. Repair typically involves
+downloading other parts of the content and using erasure coding to restore the
+missing parts. This incurs costs for a host. To compensate the host for these
+costs it receives not only its own collateral back at the end of the contract,
+but also the remaining collateral from the host that failed to hold a slot.
+
+We expect the collateral to be significantly higher than the costs of repair.
+This means that hosts in the network can benefit greatly from repairs, and they
+may prioritize repairs over filling slots in new contracts. This is intentional,
+we want the network to prioritize honoring existing contracts over starting new
+ones.
 
 Renewal
 -------
--- a/design/marketplace.md
+++ b/design/marketplace.md
@@ -62,6 +62,7 @@ including a small amount of host losses that can be tolerated.
       proof probability               # frequency at which proofs are required
       proof parameters                # proof of retrievability parameters
       erasure coding                  # erasure coding parameters
+      dispersal                       # dispersal parameter
 
       hosts                           # amount of storage hosts (including loss)
       loss                            # number of allowed host losses
@@ -184,6 +185,42 @@ of the contract, it allows people to easily chip in to host content, etc.) it
 has one big disadvantage: hosts no longer know for how long they'll be kept to
 the contract. When a contract is continuously topped up, they cannot leave the
 contract without losing their collateral.
+
+Dispersal
+---------
+
+For a client it is beneficial when their content is stored on as many different
+hosts as possible, to guard against host failures. Should a single host fill all
+slots in the contract, then the failure of this single host could mean that the
+content is lost. On a network level, we also want to avoid that a few large
+players are able to fill most contract slots, which would mean that the network
+becomes fairly centralized. Therefore each contract includes a dispersal
+parameter that helps spread content over many hosts and avoid centralization in
+the network.
+
+The dispersal parameter allows a client to choose the amount of spreading within
+the network. When a slot becomes empty then only a small amount of hosts in the
+network are allowed to fill the slot. Over time, more and more hosts will be
+allowed to fill a slot. Each slot starts with a different set of allowed hosts.
+
+The speed at which new hosts are included is chosen by the client. When the
+client choses a high speed, then very quickly every host in the network will be
+able to fill slots. This increases the chances of a single host to fill all
+slots in a contract. When the client choses a low speed, then it is more likely
+that different hosts fill the slots.
+
+We use the Kademlia distance function to indicate which hosts are allowed to
+fill a slot.
+
+    distance between a and b:   xor(a, b)
+    slot start point:           hash(nonce || slot number)
+    allowed distance:           elapsed time * dispersal parameter
+
+A host is allowed to fill a slot when the distance between its id and the start
+point is less that the allowed distance.
+
+Note that even though we use the Kademlia distance function, this bears no
+relation to the DHT. We use the blockchain address of the host, not its peer id.
 
 Conclusion
 ----------
--- a/design/marketplace.md
+++ b/design/marketplace.md
@@ -99,11 +99,16 @@ The time interval that a slot is filled by a host determines the host payout;
 for every second of the interval a certain amount of tokens are awarded to the
 host. Hosts that fill a slot are required to submit frequent proofs of storage.
 
-When a proof is missed, the collateral associated with a slot is used to pay a
-fee to the one who marked the proof as missing. The rest of the slot collateral
-is reserved for repairs. The slot is now considered empty again until another
-host submits a correct proof together with collateral. Payouts for the time
-interval that a slot is empty are burned.
+When a certain number of proofs is missed, the slot is considered empty again.
+The collateral associated with the slot is used to pay a fee to the node that
+indicated that proofs were missing. The rest of the slot collateral is reserved
+for repairs. An empty slot can be filled again once another host submits a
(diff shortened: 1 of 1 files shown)


design/marketplace.md
```diff
@@ -0,0 +1,219 @@
...
+except for a fee for the one who marked the proof as missing. The slot is now
+considered empty again until another host submits a correct proof together with
+collateral.
+
+To incentivize repair, we allow the host that takes over a slot to count the
+interval between the missed proof and the new proof towards its own profits.
+This means that when a proof is missed, the incentive to repair increases over
+time.
+
+The situation where the payout on repair becomes larger than the required
```
dryajov, line 111: This might also introduce other undesired behaviors, for example, nodes might artificially wait before filling a slot. I would proposed we avoid this by burning the payout for each missing proof, that way the only way to maximize profit is to occupy the slot as soon as possible.

dryajov:
This is great @markspanbroek, a very simple and intuitive design! I like it very much. 

Just a few comments/questions.

- When does the contract actually considered started, as in when can the client leave the network and be assured that the data will be kept safe by the network? 
  - I have some ideas here, but I'd like to hear yours first :)
- A general observation with incentives. As I mentioned in https://github.com/status-im/codex-research/pull/83#pullrequestreview-986615148, this could easily lead to some undesired behavior. I usually like to think of this as "compensation should be proportional to resources expended" - if a node isn't expending any resources and is maximizing profit by simply waiting, it can easily be abused. On the other hand, repair does incur in additional expenses, both on the remaining nodes in the network, as well as on the taking over the slot, so perhaps there should be something in between, where we calculate how long would it take the new node to rebuild the missing pieces and how many proofs on average it would have missed and amortize them to the overall payout of the node?

emizzle:
> so perhaps there should be something in between

Brainstorming here. Maybe one possible way to incentivise repair while disincentivising waiting, would be to issue a small amount of collateral from the exited slot's host (where most of their collateral is burned minus the fee to the finder who marked the proof as missing) to the host that takes over the slot? The amount of collateral retained, and passed to the new host, would have to be proportional to the computation performed, such that it does not incentivise hosts to stop

emizzle:
@bulatz wrote (including here for posterity):
>> When too many hosts fail to submit storage proofs, and no other hosts take over the slots that they vacate, then the content can be considered lost. The contract is considered failed. The collateral of every host in the contract is burned

> I feel it's unfair - why I lose my collateral if other, independent hosts failed? what I can do?

I agree with this. If I'm a host and I have acted in accordance with my responsibilities on the network, why should there be a situation in which I lose my collateral? Is there a way that a host could act in a way to prevent this from happening?

design/marketplace.md
markspanbroek: I've been struggling with this a bit. Hosts need to expend extra resources to repair a file (downloading from other hosts and paying for their bandwidth, performing the erasure coding operations), which would indicate that some sort of compensation is required. But that compensation should not become too big to avoid tampering with other incentives.

Could you elaborate on the scenario where nodes artificially wait before filling a slot? I don't see a direct problem there; if they are interested in filling a slot, then it's in their best interest to fill the slot as soon as possible, lest someone else beats them to it.

markspanbroek:
> * When does the contract actually considered started, as in when can the client leave the network and be assured that the data will be kept safe by the network?

The contract is considered started when all slots are filled. If the slots do not fill before the timeout, then the contract is considered cancelled. The timeout is chosen by the client. So the client can come back and check the state of the network after the timeout and be assured that either the data is kept safe by the network, or it will have to post a new contract (with higher payout for instance).

markspanbroek:
> feel it's unfair - why I lose my collateral if other, independent hosts failed? what I can do?

I actually agree. It is unfair to the hosts. But what are the alternatives? If we do not punish these nodes, then they can decide to wait for the contract to fail so they can drop out instead of helping with repair.

emizzle:
> However, it might be difficult to determine the amount that should be reserved for repair; network and market conditions can fluctuate, leading to different costs associated with repair. How can we ensure that we have enough money for repair?

Really great point. Further to the question, if the amount of collateral that is to be reserved for repair is dynamic based on market and network conditions, the act of estimating the cost would itself incur a cost as the network would need to be traversed and bandwidth would have be to be paid for. We'd likely need to estimate quite a bit. Have we defined a way to determine the cost of bandwidth yet? Possibly we could borrow some of the assumptions from this?

emizzle:
> I actually agree. It is unfair to the hosts. But what are the alternatives? If we do not punish these nodes, then they can decide to wait for the contract to fail so they can drop out instead of helping with repair.

If repair is incentivised properly, wouldn't nodes be more inclined to help with repair instead of waiting for the contract to fail? It should not be in their interest for a contract to fail, so perhaps we could include some kind of incentive that is distributed at the conclusion of a contract, incentivising hosts to remain til the contract's completion?

markspanbroek:
> Have we defined a way to determine the cost of bandwidth yet?

Currently nodes can set their own price for bandwidth; the market determines the price.

emizzle:
> Currently nodes can set their own price for bandwidth; the market determines the price.

That makes sense. In the case of repair, the market value for bandwidth has been determined for that contract and so that value could continue to be used to estimate the repair process. Or, perhaps a repair percentage of collateral could be one of the values in the contract that is agreed to by hosts who enter in to the contract.

emizzle:
> This would definitely work. But it also shifts the burden from the protocol designers to the client who now has to determine a correct value when starting the contract :)

We could provide a default value, and upper and lower bounds 😃  Maybe there will be some metrics to leverage on an active network that could help determine a best price before the contract is created. 

Alternatively, we could give the remaining collateral (all of the collateral minus the fee to the finder, and perhaps a fee to the codex network fund 😄 ) to the host who does the repair and fills the slot, but delay the transfer of that collateral until the completion of the contract, incentivising the node to stay in the game until the contract is finished. The host would nearly get free participation in the contract as long as they see it til the end.

markspanbroek:
> If repair is incentivised properly, wouldn't nodes be more inclined to help with repair instead of waiting for the contract to fail? It should not be in their interest for a contract to fail, so perhaps we could include some kind of incentive that is distributed at the conclusion of a contract, incentivising hosts to remain til the contract's completion?

The scenario that I have in my head when thinking about this is the following. A host has agreed to be in a contract, but after a while more lucrative contracts start to appear on the network. The contract still has a long running time, and the only thing keeping the hosts from ditching the contract is the loss of collateral.

In this scenario, anything that allows the host to drop out of the contract without losing collateral is a win for the host, including trying to frustrate repairs.



> we could give the remaining collateral (all of the collateral minus the fee to the finder, and perhaps a fee to the codex network fund smile ) to the host who does the repair and fills the slot, but delay the transfer of that collateral until the completion of the contract, incentivising the node to stay in the game until the contract is finished. The host would nearly get free participation in the contract as long as they see it til the end.

That's a really interesting idea. A sort of "double or nothing" collateral. I have think this one through a bit more.

design/marketplace.md
leobago: I also agree that in the way this part is currently written, one might think that waiting could be a good strategy to maximize profit, but it is also true that nodes are in competition, so it is in their best interest to fill the slot before others. 
It is clear that data repairing is more costly than initial download so providing a few extra incentive might be fair.

leobago:
Another thing we might want to take into account would be to load balance the workload across all storage nodes, instead of having always the same X nodes taking most of the contracts. This is important for data dispersal, but also for repairs (easier to repair 100GB of data than 10TB), as well as to distribute the load on PoR. I don't know if the marketplace should take care of this but we should try to avoid the scenario where the same nodes are always the firsts to fill the contract slots.

Menduist:
Nice proposal!

My main worry with it is the fact that no one is "elected" to download a specific slice. Meaning that when I post this contract, a lot of storage nodes are going to request the (same) blocks from me, until one of them is lucky enough to be the first one to have the full slice, and able to take the slot

It seems very wasteful, and more worryingly, if there are enough storage nodes in the network, you'll effectively get DDoS when you put up a contract
Also, as a storage node, unless I'm the fastest one, I'll have a lot of trouble to get a slot in the contract, and will effectively download a lot of blocks for nothing

I don't see a good way to stop this behavior. Maybe the slots are small enough that it doesn't matter, eg, 1 block? But then, the contract gets big and costly.
Maybe once you have one block of a slice, you can reserve the slot with a single proof, and as time goes on, you'll have to prove more and more blocks. But that's seems sketchy

-------
Other issue I see, which is not new, but worsen by this proposal, is the fact that, as a storage node, I don't have any incentive to "share" a contract.
Let's say I take all the slots for myself of a 100 slots contract. It means that if I loose my disk, I'll loose my stake for 100 slots
But effectively, this data will be lost

Now let's say I take 1 slot on 100 contract. If I loose my disk, I'll again loose my stake for 100 slots, but the data will still be there. But I don't really care
My point is, maybe we should not burn all the stake if you loose your slice of data. Instead, maybe we burn 50%*
However, if the full dataset is lost, we burn everything. This incentivize storage node to diversify their storage contracts

*: 50% may seem too low. But we should take some lessons from eth2, where they incentivized client diversity by making penalties bigger when a lot of nodes are down. And people don't seem to care too much about it

cskiraly:
> Nice proposal! I really like the direction
> 
> My main worry with it is the fact that no one is "elected" to download a specific slice. Meaning that when I post this contract, a lot of storage nodes are going to request the (same) blocks from me, until one of them is lucky enough to be the first one to have the full slice, and able to take the slot
> 
> It seems very wasteful, and more worryingly, if there are enough storage nodes in the network, you'll effectively get DDoS when you put up a contract Also, as a storage node, unless I'm the fastest one, I'll have a lot of trouble to get a slot in the contract, and will effectively download a lot of blocks for nothing
> 
> I don't see a good way to stop this behavior. Maybe the slots are small enough that it doesn't matter, eg, 1 block? But then, the contract gets big and costly. Maybe once you have one block of a slice, you can reserve the slot with a single proof, and as time goes on, you'll have to prove more and more blocks. But that's seems sketchy

Nice proposal indeed!

I share this concern about resource utilization and lack of allocation. For me a fundamental question is: Why would a storage node ask for a specific piece? Why would it matter for that node which piece it receives and stores? 
Of course, if we consider also the erasure coding, there are differences between systematic and parity blocks, but other than that it should not really matter, and with interleaving we can solve that asymmetry.
So what if the client allocates pieces on its own, with a strategy of its choice based on incoming request, doing FCFS, or batching and selecting random, etc.?

markspanbroek:
> Another thing we might want to take into account would be to load balance the workload across all storage nodes, instead of having always the same X nodes taking most of the contracts. 

This is a good point @leobago. Ideally the protocol encourages dispersal over different nodes. Since we can't really tell whether nodes are sharing a disk, it would be nice to have an incentive to not have too many eggs in one basket. I thought that the collateral would take care of that, but then:

> Other issue I see, which is not new, but worsen by this proposal, is the fact that, as a storage node, I don't have any incentive to "share" a contract.

As @Menduist points out, the node loses the same amount of collateral on disk failure, whether it takes all slots in a single contract, or several slots in several contracts. I do like the solution of burning half of the collateral on slot failure, and all of the collateral on contract failure.

emizzle:
As storage nodes fill the slots in the contract, could they be randomly assigned a slot position? So when a repair situation occurs, the contract can choose which host can do the repair by starting at index 0 and going to the end of all hosts in slots. The next repair that comes up, start at index 1, and so on. The order of precedence on who gets to do the repair is effectively random, and each host gets their turn in time.

emizzle:
Without derailing the current conversation too much, I was thinking a bit about our dynamic data future -- what happens in a  "private drive" use case when a client wants to delete a file or replace a file with one of a different size? To allow for this type of behaviour, perhaps the storage contract could include enough slots to satisfy a storage limit, eg 500 GB. Because the client may not use up all the space at first, some slots may be empty (no data), however the client would still pay for the host to occupy the slot even though they aren't really storing any data, but they are being paid for remaining idle and ready to store data. It's a bit like paying a monthly fee for a certain amount of cloud storage space. From a client's point of view, they are paying a fee to have the ability to store up to a certain amount of data in the network.

From how I understood it, the proposal depends on the content to be erasure coded *before* the number of slots in a contract can be determined, right? That perhaps might make the data too permanent for the duration of the contract. Instead, perhaps the number of slots in a contract could be defined by desired total size of storage, eg 500 GB. The tradeoff would be that it takes longer to start a contract as there would be more hosts needed to fill the slots. 

I do understand dynamic data is not a problem to solve today, but it may be interesting to inform our design decisions today.

markspanbroek:
> As storage nodes fill the slots in the contract, could they be randomly assigned a slot position? So when a repair situation occurs, the contract can choose which host can do the repair by starting at index 0 and going to the end of all hosts in slots. The next repair that comes up, start at index 1, and so on. The order of precedence on who gets to do the repair is effectively random, and each host gets their turn in time.

I didn't want to constrain repair to the hosts that already host other parts of the contract. Any host on the network should be able to repair and fill an open slot.

cskiraly:
> > So what if the client allocates pieces on its own, with a strategy of its choice based on incoming request, doing FCFS, or batching and selecting random, etc.?
> 
> I thought about this, but besides the points that Mark mentioned, it seems too easy to manipulate If I want to stop you from creating your contract, I can create a lot of "shallow" block requests, and throw the block away instantly If I'm able to create the majority of request you're receiving, actual storage node will never be able to start the contract

Still, I'm not sure you are better off with a download and proof based method. Can't someone DDoS that with the same strategy?

> Also, it doesn't work if the file is already on the network (renewal or other)

Good point.

markspanbroek:
Thanks for all the great comments so far! 

I updated the proposal to include your ideas around the repair incentive and the strategy for assigning hosts to slots.

design/marketplace.md
```diff
@@ -0,0 +1,279 @@
...
+            ---------------- time ---------------->
+
+
+The time interval that a slot is filled by a host determines the host payout;
+for every second of the interval a certain amount of tokens are awarded to the
+host. Hosts that fill a slot are required to submit frequent proofs of storage.
+
+When a proof is missed, the collateral associated with a slot is used to pay a
+fee to the one who marked the proof as missing. The rest of the slot collateral
+is reserved for repairs. The slot is now considered empty again until another
```
emizzle, line 104: This feels a bit harsh. Maybe we could allow for a few missed proofs before the the host loses their collateral and the slot is freed up?

design/marketplace.md
```diff
@@ -0,0 +1,279 @@
...
+When a contract is about to end, and someone in the network wants the contract
+to continue for longer, then they can post a new contract with the same content
+identifier.
+
+We've chosen not to allow top-ups of existing contracts with new funds. Even
+though this has many advantages (it's a very simple way to extend the lifetime
+of the contract, it allows people to easily chip in to host content, etc.) it
+has one big disadvantage: hosts no longer know for how long they'll be kept to
+the contract. When a contract is continuously topped up, they cannot leave the
+contract without losing their collateral.
```
emizzle, line 187: Renewal feels like it would be something that is a very nice to have from a user experience perspective. Not allowing for renewable contracts prevents hosts from being perpetually locked in to contracts against their will. However, couldn't there be a property in the contract that a host can set allowing them to opt in or out of renewals?

design/marketplace.md
```diff
@@ -0,0 +1,279 @@
...
+        v
+    ---------
+    | ended |
+    ---------
+
+
+Repairs
+-------
+
+When a slot becomes empty, the remaining collateral associated with the slot is
```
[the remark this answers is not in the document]
markspanbroek: I'm all for getting this amount to be as close as possible to the actual costs of repair. I just don't know any simple way of approximating these costs.

The download receipts sound complex to me for several reasons:
1. They introduce a tight coupling between the storage contracts and the bandwidth incentives. 
2. Payments are handled off-chain in channels, contracts are handled on-chain, which means we have to figure out how to present these payment on-chain.
3.  It seems easy to artificially inflate the price of downloads using a sybil attack whereby you sell the data to yourself. Preventing this seems non-trivial.

Perhaps a fixed price for repair that is set at contract start is at the sweet spot of being both simple and effective enough?

design/marketplace.md
markspanbroek: Yes, I agree. I will change the wording here to indicate that a number of proofs need to be missing before triggering repair.

design/marketplace.md
markspanbroek: We can still offer renewal to the user; the node would simple posts a new contract with the same parameters.

design/marketplace.md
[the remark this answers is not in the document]
Menduist: Actually, it makes even more sense in that new design. Otherwise, hosts can set their bandwidth price very high, and since the client doesn't choose the host, they could hold the data hostage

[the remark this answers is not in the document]
dryajov: > I'm all for getting this amount to be as close as possible to the actual costs of repair. I just don't know any simple way of approximating these costs.

This should be doable. We can quantify both the bandwidth and repair costs and we've been looking at this to some extent already - @cskiraly.

> The download receipts sound complex to me for several reasons:
> They introduce a tight coupling between the storage contracts and the bandwidth incentives.
> Payments are handled off-chain in channels, contracts are handled on-chain, which means we have to figure out how to present these payment on-chain.

That should be fine? Ultimately, payment channels are settled on chain and the receipts would be verifiable on-chain as well. It should be just a call to the payment channel contract?

> It seems easy to artificially inflate the price of downloads using a sybil attack whereby you sell the data to yourself. Preventing this seems non-trivial.

Not sure what a sybil attack has to do with this? Every entity is bounded by a stake. An entity attempting to subvert the system by creating many identities exposes itself to (at least) a proportional amount of risk to the number of entities it brings to the system.

Never the less, it's a good point and perhaps you are right that we should set a fixed price for repair.

> Perhaps a fixed price for repair that is set at contract start is at the sweet spot of being both simple and effective enough?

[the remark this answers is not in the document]
dryajov: > If I may ramble about my old ideas, setting a maximum retrieval price in the contract would, amongst other benefits, allow us to calculate easily how much the repair costed in bandwidth

This doesn't sounds like a bad idea and I think it makes sense, but then I also like the idea of simply fixing the repair price to some average bandwidth cost + resource cost. It simplifies things a lot. 

Furthermore, I think in the case of repair, fixing the price is more acceptable because of several reasons - a) the repair node is going to further benefit from earning on the contract and b) it will also potentially benefit from download fees - so, there is a natural incentive to do the repair.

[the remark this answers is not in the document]
dryajov: There are two scenarios to watch out for here.

1. If download prices can be inflated, then nodes will attempt to collude to collect the outstanding collateral
2. If the price of repair is much larger than the potential benefits from storing the data, then no one is going to attempt it

So, whatever we choose to do, it should be somewhere in between those two extremes - nodes should not be able to collect the entire collateral during repair and the repair process should not be more expensive than the rewards from storing the data.

Again, a fixed price for repair that averages the costs for bandwidth + resources doesn't sound like a bad idea.

The more interesting issue is what happens if nodes choose not to participate in the repair? I tend to look at this as follows:

- If the redundancy levels are close to optimal, then repair can be lazy and at the same time there should be enough nodes, with enough pieces sufficient to do the repair
- If redundancy levels decrease and are getting close to `r0`, then nodes are incentivized to actually help with repair as much as possible, otherwise, once data is lost, everybody looses the stake

`r0` is a per dataset parameter that is somewhere in between `K` - the minimum number of pieces required to reconstruct and `N` - all the available pieces.

dryajov:
I've captured some hight level requirements here #88

design/marketplace.md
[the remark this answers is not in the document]
markspanbroek: Ok, all things considered I think that having a fixed price for repairs is the optimal solution for now. I changed the wording in the proposal to reflect this.

markspanbroek:
Merging this, so that we can have new pull requests with changes to design.

PR #85, opened 2022-05-28
erasure coding design overview

dryajov:


design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
...
+2) Take $K$, $S$ spaced blocks and $K$ symbols $s$ from each of the selected blocks at offset 0 and encode into $M$ parity symbols, each placed in new $S$ spaced blocks at the same offset and appended to the end of the sequence of original blocks. Repeat this steps at offset $1*sizeof(GF(p))$ and so on.
+3) Repeat the above steps for the length of the entire dataset.
+
+Bellow is a graphical outline of the process:
+
+_The sequence of original blocks_
+
+![](./figs/blocks.svg)
+
+_The logical matrix resulting from stacking each $S$ symbols together, where $K=3$ and $S=4$_
```
leobago, line 47: The math rendering is not working well here.

design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
...
+
+![](./figs/matrix1.svg)
+
+_Symbols and Coding direction. Each cell corresponds to a  symbol $s$ and each column of symbols are coded together_
+
+![](./figs/matrix3.svg)
```
leobago, line 53: In this figure S=4 instead of S=3 as in the other figures.

design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Load balancing retrieval
+
+There is an issue resulting mostly from the systematic nature of the code and the presence of original and parity data.
+
+It is safe to assume that the systematic (original) data is going to be retrieved more frequently than the parity data, which should only be retrieved when one of the systematic pieces have gone missing thus, in order to prevent congestion and overloading of the systematic nodes some further placement considerations should be taken in to account. Namely, having nodes dedicated to parity data should be avoided.
```
leobago, line 73: one of the systematic pieces has gone missing..

cskiraly:
I still think the ordering, at the block level (not at the symbol level) should be the other way around, although admittedly it only matters if one considers partial access and sequential data access patterns. In that case it allows for the parallel acceleration of sequential reads that you would not have with this ordering.

design/erasure-coding.md
dryajov: Not sure I understand what the proposed change is?

cskiraly:
> > I still think the ordering, at the block level (not at the symbol level) should be the other way around, although admittedly it only matters if one considers partial access and sequential data access patterns. In that case it allows for the parallel acceleration of sequential reads that you would not have with this ordering.
> 
> What do you mean by the other way around? Not sure if I follow :)

What I mean is that the original block order (1, 2, 3, etc.) goes vertically in the grid representation, forming protection and also acceleration over adjacent blocks. This with the assumption that original block order actually matters for those accessing it, i.e. it is a typical access pattern.

dryajov:
> > > I still think the ordering, at the block level (not at the symbol level) should be the other way around, although admittedly it only matters if one considers partial access and sequential data access patterns. In that case it allows for the parallel acceleration of sequential reads that you would not have with this ordering.
> > 
> > 
> > What do you mean by the other way around? Not sure if I follow :)
> 
> What I mean is that the original block order (1, 2, 3, etc.) goes vertically in the grid representation, forming protection and also acceleration over adjacent blocks. This with the assumption that original block order actually matters for those accessing it, i.e. it is a typical access pattern.

I see, yes this can make sense but it heavily depends on the placement we choose. I think this needs to be further clarified in the text and worded in such a way that it is clear that placement considerations would affect interleaving and vice-versa.

leobago:
If we consider that a slot is a row (row: blocks belonging to different encoding columns to disperse erasures) then the ordering proposed by @dryajov will be beneficial for those reading sequential sections of the dataset. The other configuration (having sequential blocks 1, 2, 3.. in a column) would only provide acceleration benefits during the encoding. Assuming we encode once and read several times, the ordering proposed by @dryajov seems more beneficial to me.

Bulat-Ziganshin:
We may have several different layouts. F.e. let's specify one of them, a simplified one with fair distribution of download bandwidth:

User provides `S*K` blocks and requests K+M encoding saving data over K+M nodes (thus each node stores S blocks).

For the first K nodes, block number Y on node X = original block number `Y*K+X` (i.e. blocks 0, 1 .. K-1, K... goes into nodes 0, 1 .. K-1, 0...). In other words, original block number I goes into block number I/K on node I%K.

For the last M nodes, block number Y on node X is a part of recovery group consisting from blocks number Y on all K+M nodes, i.e. original blocks `Y*K+i, i=0..K-1` plus M recovery blocks.

Note however that this scheme places all original data to the first K nodes and all recovery data to the last M nodes. If we want to spread recovery data fairly over all nodes, we should reshuffle data between servers in each recovery group (note that a single recovery group still occupies block Y on all K+M severs).

The simplest way to do that is to shift every next group by one position relative to previous one:
- blocks `0 .. K-1` are mapped to servers `0 .. K-1`, correspondingly
- block `K .. 2*K-1` are mapped to servers `1 .. K`, correspondingly
- block `2*K .. 3*K-1` are mapped to servers `2 .. K+1`, correspondingly
and so on.

Recovery blocks are shifted accordingly, filling remaining servers in each recovery group:
- recovery blocks `0 .. M-1` (i.e. recovering data from original blocks 0..K-1) are mapped to servers `K .. K+M-1`, correspondingly
- recovery blocks `M .. 2*M-1` are mapped to servers `K+1 .. K+M-1, 0` correspondingly
- recovery blocks `2*M .. 3*M-1` are mapped to servers `K+2 .. K+M-1, 0, 1` correspondingly

Bulat-Ziganshin:
That's a writeup of arguments for using "layout B", i.e. layout described in my previous message, trying to put it in the form suitable for inclusion in the document.

### Preconditions

We assume that:
- The main source of large data losses are situations when an entire host drops (temporarily or permanently) or a host loses an entire hdd/ssd
  - Thus, each recovery group should include as much hosts as it possible without degrading other characteristics
- The dataset, that provided by user, has some internal structure and sequential data blocks are often more closely related than more distant blocks, f.e. they may be parts of the same file or closely related files, if the dataset is an archive or a disk image
  - Thus, reading of R sequential blocks is more likely than reading of other, arbitrary chosen sets of R blocks
  - Also, if we are going to lose subset of data, it's probably better to lose R sequential blocks (i.e. one file entirely) than arbitrary R blocks (i.e. one block in each of R files)
- The download requests should be fairly distributed among all hosts


### Simple layout (reasoning and textual description)

The setting:
- User selects recovery groups consisting of N=K+M blocks where K blocks represent the original data and M blocks are the computed ECC data
- The entire dataset provided by user consists of S such groups
- User selects the number of storage hosts H, and amount of data stored on the each host

We propose here layout only for the simple case of H=N hosts, each storing S blocks. More complex layouts are TBD.

Based on assumptions above, we choose the following layout:
- In order to maximize recovery chances, we distribute each recovery group over all N hosts. This allows to recover all data even if up to M hosts are offline for any reason
- In order to maximize sequential read throughput, we combine each group from K sequential original blocks
- In order to evenly distribute download requests between all hosts, we evenly spread ECC blocks over hosts
- In order to max out sequential read throughput, we ensure that if original block R is placed on the host Q, then block R+1 is always placed on host `(Q+1) mod N` (round-robin placement)


### Simple layout (strict math definition)

The algorithm describes how user-provided data are emplaced to hosts, and ECC data are generated from them and emplaced too:
- On input, we have `S*K` user-provided data blocks `INPUT[I], I=0..S*K-1`
- On output, we have N nodes each storing S blocks `OUTPUT[I][J], I=0..N-1, J=0..S-1`

First, let's organize data into recovery groups and generate ECC:
- `GROUPED[I][J] = INPUT[I+J*K], I=0..K-1, J=0..S-1` are the original data
- `GROUPED[K .. K+M-1][J] = ECC(GROUPED[0..K-1][J]), J=0..S-1` are the computed ECC data - M recovery blocks computed from K original blocks of the same recovery group

Now, the asymmetric block placement (first K nodes contains all data blocks):
- `OUTPUT[I][J] = GROUPED[I][J], I=0..N-1, J=0..S-1`

And symmetric, round-robin block placement (data and ECC blocks are evenly distributed between nodes):
- `OUTPUT[(I+K*J) mod N][J] = GROUPED[I][J], I=0..N-1, J=0..S-1`

dryajov:
@Bulat-Ziganshin looks good!

> Also, if we are going to lose subset of data, it's probably better to lose R sequential blocks (i.e. one file entirely) than arbitrary R blocks (i.e. one block in each of R files)

Just to clarify, we're always treating a dataset as a single unit, this means that there isn't any "preference" in which blocks are lost, except for preventing entire columns from being lost.

Also, to clarify on the terminology, what precisely do you mean by "recovery group"? As far as I can infer, it would be what we refer to the "column" in the original writeup, i.e. an entire codeword?

dryajov:
@leobago the graphic seems to be inverted, `K` would be the column and `S` would be the row, consequently, `M` is also stacked in the column and additional parity columns are simply appended to the original dataset, which end up in the `S` dimension. This doesn't change with the reshuffling proposed by @Bulat-Ziganshin.

Bulat-Ziganshin:
@dryajov proposed an alternative design which allows to append ECC blocks to the datastore, simplifying indexing and adding recovery on top of pure data or replacing recovery scheme with a different one.

I will show it as example first. Let's we have 3 groups of 5 blocks each and want to add 2 parity blocks to each group. So original blocks are 0..14 and parity blocks will be 15..20. Their layout (each row is a single node) on 5+2=7 nodes will be:

```
0  7 14
1  8 15
2  9 16
3 10 17
4 11 18
5 12 19
6 13 20
```

Dmitry's idea is to use diagonales as recovery groups. I.e. the first recovery group will consist of blocks 0 8 16 3 11 19 6, second one of blocks 7 15 2 10 18 5 13, and last one of blocks 14 1 9 17 4 12 20. I.e. group I starts at I'th block at the server #0 and includes block (I+J) mod S on server #J.

Bulat-Ziganshin:
The "layout B" has advantage for sequential reading when some nodes lost: if we have M+K encoding, then we can receive one block per node from any M nodes and perfrom a single decoding operation to recover M sequential data blocks (because they belongs to the single recovery group). In layout A and layout C, sequential M blocks are distributed among different recovery groups.

F.e. in the picture below we can ready any 4 blocks of [1, 2, 3, 4, P1, P2] and recover first 4 blocks of the file using a single decode operation. With layout A or C, it requires to read more data and perform multiple decoding operations.

image

dryajov:
> The "layout B" 

I'm not sure if I follow which layout is which at this point :). It would be helpful if we can write a short summary of each for reference so we can all follow, maybe we can even do this in a separate write (in discussions) up and reference it here.

design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
...
+
+The size of the codeword determines the maximum number of symbols that an error correcting code can encode and decode as a unit. For example, a Reed-Solomon code that uses a Galois field of $2^8$ can only encode and decode 256 symbols together. In other words, the size of the field imposes a natural limit on the size of the data that can be coded together.
+
+Why not use a larger field? The limitations are mostly practical, either memory or CPU bound, or both. For example, most implementation rely on one or more arithmetic tables which have a linear storage overhead proportional to the size of the field. In other words, a fields of $2^{32}$ would require generating and storing several 4GB tables and it would still only allow coding 4GB of data at a time, clearly this isn't enough when the average high definition video file is several times larger than that and specially when the expectation is to allow handling very large, potentially terabyte size datasets, routinely employed in science and big-data.
+
+## Interleaving
```
leobago, line 22: I think this whole section needs to be better explained, it is not clear to me what you are trying to say here.

design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
...
+
+Interleaving is the process of combining or interweaving several symbols from disparate parts of the source data in ways that minimizes the likelihood that any sequence of these symbols damaged together compromises the rest of the data.
+
+In our case, the primary requirement is to ensure that we're preserving the dataset in its entirety and any $K$ elements are still enough to recover the original dataset. This is important to emphasize, we cannot simply encode a chunk of data individually, that would only protect that individual chunk, the code should protect the dataset in its entirety, even if it is terabytes size.
+
+A secondary but important requirement is to allow appending data without having to re-encode the entire dataset.
+
+### Codex Interleaving
```
leobago, line 30: This section is nicely explained

design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
...
+
+The resulting matrix with parity blocks added, where $K=3$, $M=2$, $N=5$ and $S=4$
+
+![](./figs/matrix2.svg)
+
+The resulting structure is a matrix of height $N$ and width $S$. It allows loosing any $M$ "rows" of the matrix and still reconstructing the original dataset. Of course, loosing more than $M$ symbols in a given column would still render the entire dataset invalid, but the likelihood of that happening is mitigated by placing the individual original and parity blocks on independent locations.
+
+Moreover, the resulting dataset is still systematic, which means that the original blocks are unchanged and can be used without prior decoding.
+
+## Data placement and erasures
```
leobago, line 63: Nicely explained as well

design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
...
+
+The code introduced in the above section satisfies our original requirement of any $K$ blocks allowing to reconstruct the original dataset. In fact, one can easily see that every row in the resulting matrix is a standalone element and any $K$ rows will allow reconstructing the original dataset.
+
+However, the code is $K$ rows strong only if we operate under the assumption that each failure is independent of each other. Thus, it is a requirement that each row is placed independently. Moreover, the overall strength of the code decreases based on the number of dependent failures. If we place $N$ rows on two independent locations, we can only tolerate $M=N/2$ failures, three will allow tolerating $M=N/3$ failures and so on. Hence, the code is only as strong as the number of independent locations each element is stored on. Thus, each row and better yet, each element (block) of the matrix should be stored independently and in a pattern mitigating dependence, meaning that placing elements of the same column together should be avoided.
+
+### Load balancing retrieval
```
leobago, line 69: Very clear section. Maybe we can extend it by exploring some placing strategies we discussed with Bulat.

design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
...
+
+There is an issue resulting mostly from the systematic nature of the code and the presence of original and parity data.
+
+It is safe to assume that the systematic (original) data is going to be retrieved more frequently than the parity data, which should only be retrieved when some of the systematic pieces have gone missing thus, in order to prevent congestion and overloading of the systematic nodes some further placement considerations should be taken in to account. Namely, having nodes dedicated to parity data should be avoided.
+
+This can be accomplished by choosing a placement pattern that avoids favoring some nodes over others when it comes to storing the systematic data, and at the same time it should not break the placement rules around the safety of the code. For example, one can think of a placement strategy where the node gets assigned a block based on the node's position in some queue or its id or a combination thereof.
+
+Some placing strategies will be explored in a further document.
+
+### Adversarial vs random erasures
```
leobago, line 79: Good. Perhaps add a link to our PoR docs?

leobago (commented):
I was working on our Erasure Coding write-up, and I thought about checking this document, so I left some comments for improvement. I think the overall explanation is very good, except for a few parts that can be improved, I think I can take a big chunk of this and complement it with other work we have done in order to have a document ready for posting soon.

design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
...
+
+The size of the codeword determines the maximum number of symbols that an error correcting code can encode and decode as a unit. For example, a Reed-Solomon code that uses a Galois field of $2^8$ can only encode and decode 256 symbols together. In other words, the size of the field imposes a natural limit on the size of the data that can be coded together.
+
+Why not use a larger field? The limitations are mostly practical, either memory or CPU bound, or both. For example, most implementation rely on one or more arithmetic tables which have a linear storage overhead proportional to the size of the field. In other words, a fields of $2^{32}$ would require generating and storing several 4GB tables and it would still only allow coding 4GB of data at a time, clearly this isn't enough when the average high definition video file is several times larger than that and specially when the expectation is to allow handling very large, potentially terabyte size datasets, routinely employed in science and big-data.
```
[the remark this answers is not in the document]
dryajov: This is in the context of small vs large codeword codes which allow encoding large chunks of data without (as you suggest), having to split them it into smaller chunks. This is prelude/setup for the following sections which lays out the interleaving, which is the "dividing them into smaller segments"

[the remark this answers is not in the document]
dryajov: Maybe this can be further extended to clarify this context?

design/erasure-coding.md
dryajov: Yes, this isn't a simple topic to grasp and it might require a deeper explanation. I'm trying to elucidate the difference between large and small codeword sizes and why they are a limitation in the context of some GF and Reed-Solomon implementations, which effectively limit the size of the field and as a consequence the amount of data that is being encoded as a whole. 

The problem here is that most of the commonly used algorithms so far have been O(n^2), which practically limits the code to be a small codeword. However, this isn't the case for all implementations anymore and more modern algorithms allow working with much larger fields, for example FastECC and Leopard.

Understanding the meaning of codeword size is important because this is why we need to do interleaving in the first place. I'm obviously doing a bad job explaining this.

design/erasure-coding.md
```diff
@@ -0,0 +1,83 @@
+# Erasure Coding
+
+We present an interleaved erasure code that is able to encode arbitrary amounts of data despite the relatively small Reed-Solomon codeword size.
+
+## Overview
+
+Erasure coding has several complementing uses in the Codex system. It is superior to replication because it provides greater redundancy at lower storage overheads; it strengthens our remote storage auditing scheme by allowing to check only a percentage of the blocks rather than checking the entire dataset; and it helps with downloads as it allows retrieving $K$ arbitrary pieces from any nodes, thus avoiding the so called "stragglers problem".
```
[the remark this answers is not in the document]
dryajov: > Checking only a percentage of the blocks is probabilistically correct. However, to have 100% guarantee you still need to check at least K blocks, which comes back to same size of the file.

The probabilistic assumption here is negligible, which makes this practically equivalent - i.e. 99.99999999....% ~= 100%?

design/erasure-coding.md
Bulat-Ziganshin: I volunteered to write a text about the 3 placing strategies, with ultimate goal to make a post for our blog

PR #86, opened 2022-06-02
Model of the platform and its reliability level

leobago:
WIP: Initial model of the structure of the codex platform and expected reliability.


dryajov:
Can we please split this into two PRs on two branches belonging to this repos rather than a fork?

dryajov:
Can we please split this into two PRs on two branches belonging to this repos rather than a fork?

PR #121, opened 2022-09-02
create RFC for zkSNARK

jessiebroke:
Stub for an actual proposal @dryajov @cskiraly


easye:
Is this intended for the Vac RFC process or something else?

If the former, I would be happy to contribute a template.

jessiebroke:
Not yet @easye, this is just a stub for an initial design proposal that needs to be outlined in the PVD for Phase 2 planning.

PR #141, opened 2022-11-22
Interactions between Sales and Repo

markspanbroek:
Small design document capturing how the Sales module and the Repo interact. Captures ideas originally discussed between @dryajov, @emizzle and me.

### DIFF
--- a/design/sales.md
+++ b/design/sales.md
@@ -1,0 +1,47 @@
+Sales module
+============
+
+The sales module is responsible for selling a node's available storage in the
+[marketplace](./marketplace.md). In order to do so it needs to know how much
+storage is available. It also needs to be able to reserve parts of the storage,
+to make sure that it is not used for other purposes while a contract is running
+or being negotiated.
+
+    ---------------------------------------------------
+    |                                                 |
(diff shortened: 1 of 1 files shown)


design/sales.md
```diff
@@ -0,0 +1,47 @@
...
+    |                           ^         ^           |
+    ----------------------------|---------|-----------
+                                |         |
+                  availability  |         | state
+                                v         v
+                   ----------------    -----------------
+                   |     Repo     |    |   Datastore   |
+                   ----------------    -----------------
+
+The Sales module keeps track of storage reservations in its internal
```
leobago, line 29: Nice. We should keep in mind that a malicious node could try to spam the network by asking for contracts that are never fulfilled. If the storage reservation system keeps the reservation for too long, one could virtually render a number of nodes "out of storage" while in reality, they do have it but it is just reserved for contracts that will never materialize.

markspanbroek: Good point, thanks. There is a cost to posting a storage request, and then never releasing the data (a client loses some money when a contract never starts). This should be a disincentive to spam the network.

PR #165, opened 2023-06-05
Add request queue design

emizzle:
Adds a request queue design for queuing of requests in the marketplace. This designs attempts to facilitate processing of requests in a priority sorting order.

### DIFF
--- a/design/sales.md
+++ b/design/sales.md
@@ -187,13 +187,18 @@ processed contains one or more slot indices (as in the case of repair), the
 random slot should be chosen from that sample of indices. The start of the sales
 process should ensure that the random slot chosen is indeed available (slot
 state is "free") before continuing. If it is not available, the sales process
-will exit and the host will process the top request in the queue. Note: it may
-be that the top request in the queue is the same request that was just
-processed, indicating that the request still has slots to be filled, as the
-request was had not yet been removed due to being started/failed/cancelled. This
-re-processing of the same request allows for the possibility that the random
-slot index chosen had already been filled by another host, and gives the host a
-chance to fill a different slot index of the same request.
+will exit and the host will process the top request in the queue. The start of
+the sales process should also check to ensure the host is allowed to fill the
+slot, due to the [sliding window
+mechanism](https://github.com/codex-storage/codex-research/blob/master/design/marketplace.md#dispersal).
+If the host is not allowed to fill the slot, the sales process will exit and the
+host will process the top request in the queue. Note: it may be that the top
+request in the queue is the same request that was just processed, indicating
+that the request still has slots to be filled, as the request was had not yet
+been removed due to being started/failed/cancelled. This re-processing of the
+same request allows for the possibility that the random slot index chosen had
+already been filled by another host, and gives the host a chance to fill a
+different slot index of the same request.
 
 ### Implementation tips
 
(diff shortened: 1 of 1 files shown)


emizzle:
I believe we had initially discussed creating a slot queue, however after considering the case where there are hosts that agree on market conditions, but clashed in the random assignment of their slot indeces, could re-process a request in the queue if was not removed (for being filled/cancelled/failed/finished).

markspanbroek (approved):
Nice, thanks @emizzle!
Perhaps we can already mention that when repair is added, the queue will  also contain slots that have recently been freed?

design/sales.md
```diff
@@ -140,6 +140,47 @@ the state is kept on local disk by the Repo and the Datastore. How much space is
...
+Once a new request for storage is created on chain, all hosts will receive a
+contract event announcing the storage request and decide if they want to act on
+the request by matching their availabilities with the incoming request. Because
+there will be many requests being announced over time, each host will create a
+queue of matching requests, adding each new storage request to the queue.
+Requests in the queue will be sorted in the following order:
+1. Profit (descending, yet to be determined how this will be calculated)
+2. Collateral required (ascending)
+3. Time before expiry (descending)
+4. Data size (ascending)
```
AuHau, line 154: Hmm, what role does data size play in decision-making? I understand the roles of the previous three points, but I am unsure about this one. Smaller requests are better? Why? Maybe because it allows the host to more efficiently cover its availabilities? But then bigger requests will be "penalized" and might take more time to fulfill... Is that desired?

AuHau (commented):
Generally LGTM, but I have one question.

AuHau:
Also, shouldn't there be a mention of the Dispersal mechanism here? I think it is very much linked to when a request should enter the queue.

design/sales.md
emizzle: This is a good point. Larger data sets will cost more to download in both cost and time. It stands to reason that eventually dataset size would have an impact on profitability, as downloading the data will incur a cost in bandwidth fees. But as a first implementation attempt, profit will not be included as a sorting parameter (as we don't know how to calculate it yet) and dataset size could temporarily fill its shoes.

Yes, it seems the implications of considering dataset size in sorting the requests could consequently cause larger dataset requests to take more time to fill, when comparing to requests assuming that all other parameters of the request are the same. However, hopefully this is resolved once profitability can be calculated with dataset size included in that calculation.

emizzle:
> Also, shouldn't there be a mention of the Dispersal mechanism here? I think it is very much linked to when a request should enter the queue.

The dispersal of data may come into play in determining the profitability, as it would likely indicate the amount of data a host must download in order to generate a proof and fulfil its duties. This is part of the reason that we cannot determine profitability yet (the other main reason being unknown bandwidth incentives) and therefore can't be included in the first iteration of the request queue.

design/sales.md
emizzle: I've added a note regarding this in the `Sort order` section.

AuHau:
@emizzle sorry what I meant was the sliding window mechanism. I referenced it here as "Dispersal" as it called like that here: https://github.com/codex-storage/codex-research/blob/master/design/marketplace.md#dispersal

AuHau (approved):
👍

PR #167, opened 2023-06-21
Change request queue to slot queue

emizzle:
The request queue design was overly complicated as there was additional logic required to track one item with multiple slot indices. Instead, each item in the queue could be represented by a slot, and each time a request for storage is received, all slots for the request can be expanded and added to the queue.

### DIFF
--- a/design/sales.md
+++ b/design/sales.md
@@ -135,70 +135,88 @@ the state is kept on local disk by the Repo and the Datastore. How much space is
 reserved to be sold is persisted on disk by the Repo. The availabilities are
 persisted on disk by the Datastore.
 
-## Request queue
+## Slot queue
 
 Once a new request for storage is created on chain, all hosts will receive a
 contract event announcing the storage request and decide if they want to act on
 the request by matching their availabilities with the incoming request. Because
 there will be many requests being announced over time, each host will create a
-queue of matching requests, adding each new storage request to the queue.
+queue of matching request slots, adding each new storage slot to the queue.
 
-### Adding requests to the queue
+### Adding slots to the queue
 
-Requests will be added to the queue when request for storage events are received
+Slots will be added to the queue when request for storage events are received
 from the contracts. Additionally, when slots are freed, a contract event will
-also be received, and the request (and slot index) will be added to the queue.
-If the request of the freed slot already exists in the queue, the entry will be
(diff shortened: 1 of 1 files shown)


design/sales.md
```diff
@@ -135,70 +135,88 @@ the state is kept on local disk by the Repo and the Datastore. How much space is
...
 mechanism](https://github.com/codex-storage/codex-research/blob/master/design/marketplace.md#dispersal).
 If the host is not allowed to fill the slot, the sales process will exit and the
-host will process the top request in the queue. Note: it may be that the top
-request in the queue is the same request that was just processed, indicating
-that the request still has slots to be filled, as the request was had not yet
-been removed due to being started/failed/cancelled. This re-processing of the
-same request allows for the possibility that the random slot index chosen had
-already been filled by another host, and gives the host a chance to fill a
-different slot index of the same request.
+host will process the top request in the queue.
```
markspanbroek, line 208:
```suggestion
host will process the top slot in the queue.
```

design/sales.md
```diff
@@ -135,70 +135,88 @@ the state is kept on local disk by the Repo and the Datastore. How much space is
...
-the sales process should also check to ensure the host is allowed to fill the
-slot, due to the [sliding window
+to the queue.
+
+As soon as items are available in the queue, and there are workers available for
+processing, an item is popped from the queue.
+
+> NOTE: queue processing should not be resumed once availabilites have been
+added as this would require keeping track of all requests on the network while
+there are no host availabilities, which would have a large memory impact.
```
markspanbroek, line 196: I don't understand this, could you elaborate?

markspanbroek (approved):
Nice, thanks @emizzle!

AuHau (approved):
LGTM

design/sales.md
emizzle: In the previous design, it was stated:
> If the queue is not empty, but there are no availabilities, queue processing will resume once availabilites have been added.

The issue with this is that if a host has no availabilities, then it must keep a separate list of all new storage requests coming in from the network, so that when availabilities are created by the host, the backlog of network requests can be compared to availabilities to potentially be added to the slot queue.

emizzle: I've updated the note to reflect this

markspanbroek: That makes sense, thanks for clarifying!

PR #169, opened 2023-07-05
Initial design to reduce metadata overhead

dryajov:
This document proposes a change to reduce metadata overhead in manifest files as well as consolidating the slot blocks handling with the network block handling.

### DIFF
--- a/design/metadata-overhead.md
+++ b/design/metadata-overhead.md
@@ -1,0 +1,212 @@
+# Reducing Metadata Overhead
+
+Metadata plays a crucial role in any distributed or peer-to-peer (p2p) storage network. However, it often incurs significant overhead for the system. Therefore, it is important to understand the required metadata and how it should be stored, located, and transmitted.
+
+## Metadata and Manifests
+
+Codex, utilizes a metadata descriptor structure called the "manifest". A manifest is similar to a torrent file and stores various pieces of information necessary to describe a dataset.
+
+```
+  Manifest
+    rootHash         # Cid of root (tree) hash of the contained data set
+    originalBytes    # Exact size of the original (uploaded) file
+    blockSize        # Size of each contained block
+    blocks           # Array of dataset blocks Cids
+    version          # Cid version
+    hcodec           # Multihash codec
+    codec            # Data set codec
+```
+
+Additional information that describes erasure coding parameters may also be included:
+
+```
+  Manifest
+    ...
+    ecK           # Number of blocks to encode
+    ecM           # Number of resulting parity blocks
+    originalCid   # The original Cid of the dataset being erasure coded
+    originalLen   # The length of the original manifest
+```
+
+Manifests are treated as regular blocks of data, requiring no special handling by the Codex network or nodes. This means that announcing and storing manifests follows the same flow and uses the same subsystems as regular blocks. This convenience simplifies the execution flow significantly.
+
+## Manifest limitations
+
+Including block hashes in the manifest introduces significant limitations. Firstly, the size of the manifest grows linearly with the number of hashes and the size of the hashing function itself, resulting in increased overhead for storing and transmitting manifests.
+
+Overall, large manifests impose additional burden on the network in terms of storage and transmission, resulting in unnecessary overhead. For example, when retrieving a sizable file, it becomes necessary to obtain all the hashes listed in the manifest before downloading the initial block. This process can require hundreds of megabytes of data.
+
+One way to reduce the number of hashes is to increase the block size, which only partially addresses the problem. A better solution however, is to completely remove the blocks array from the manifest and instead rely on a Merkle proofs to verify the block.
+
+## Slots and verification subsystem support
+
+Besides the block hashes overhead, another reason for the change is the introduction of slots (verifiable dataset subsets) that nodes in a storage set/group store and verify. Slots require Merkle trees for verification, but otherwise are identical to the top-level dataset. Thus, storing and transmitting Merkle proofs is already a requirement for slot verification.
+
+Replacing the blocks array with a proper Merkle tree would allow using the same mechanism proposed in this document, for both the top level dataset and for slot verification, storage and transmission. This greatly simplifies integration of the verification subsystem.
+
+## Removing blocks array
+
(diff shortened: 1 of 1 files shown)


benbierens:
I like this plan. I have read the doc briefly and I'll read it again some more shortly, but I have a question for you:
Is it required that Merkle trees are binary, aka max 2 elements per node?
I understand how the leaf index as binary gives you the path from root to leaf, and how the hashes along the way are useful for inclusion proofs. But I was thinking... what if we allow 4 elements per node? the leaf index as binary would still represent the path from root to leaf except in chunks of 2 bits instead of 1. The same holds true for 3 bits and 8 elements per node, etc. An approach like this might greatly reduce the number of levels to our trees for large datasets, and I don't see a downside to this. Your thoughts please! :D

tbekas:
@benbierens I think that  2 elements per node is optimal when it comes to the size of the proofs. Think about the extreme case when you have N leaves and N elements per node. The size of the proof for each of the N leaves will be O(N).

@dryajov I like the Idea. I'm not convinced yet that we need to store the entire tree. I think we could (should) store only a merkle proof next to a block and that should suffice for all the use cases in my opinion. 

It will take only 2 times more on the disk, but I think it would reduce the complexity of the flows and possibly increase the performance of disk operations (sequential reads only).

design/metadata-overhead.md
```diff
@@ -0,0 +1,212 @@
...
+
+1. The client requests blocks from the network store, using the Merkle root and block index
+   1. Network store requests the block from the BlockExchange engine
+      1. BlockExchange checks connected peers for requested hash
+         1. If they do, the block is requested using the root hash and index (or range) of the block
+         2. Otherwise, it queries the DHT for the requested root hash
+            1. Once new peers have been discovered and connected, go to step 1.1.1
+2. Once blocks are received from the remote nodes
+   1. The hashes are verified against the requested Merkle root and if they pass
+      1. The block is persisted to the network
```
tbekas, line 210: Probably you meant `persisted to repo/local store`

design/metadata-overhead.md
```diff
@@ -0,0 +1,212 @@
+# Reducing Metadata Overhead
+
+Metadata plays a crucial role in any distributed or peer-to-peer (p2p) storage network. However, it often incurs significant overhead for the system. Therefore, it is important to understand the required metadata and how it should be stored, located, and transmitted.
+
+## Metadata and Manifests
+
+Codex, utilizes a metadata descriptor structure called the "manifest". A manifest is similar to a torrent file and stores various pieces of information necessary to describe a dataset.
```
markspanbroek, line 7:
```suggestion
Codex utilizes a metadata descriptor structure called the "manifest". A manifest is similar to a torrent file and stores various pieces of information necessary to describe a dataset.
```

design/metadata-overhead.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Merkle Tree
+
+The block hashes array is replaced by a Merkle tree. The Merkle tree should support persisting to disk, partial and non blocking reads/writes, loading and storing from (async) iterators. For reference, checkout out https://github.com/filecoin-project/merkletree.
+
+### Block retrieval
+
+#### Block Exchange Engine
+
+The block exchange engine requires support for querying blocks by their index and respective dataset Merkle root. It also requires returning the Merkle proofs along side the chunk so that it can be readilly verified. Scheduling blocks for retrieval should largely remain the same, but additional request and response messages are required.
```
markspanbroek, line 63:
```suggestion
The block exchange engine requires support for querying blocks by their index and respective dataset Merkle root. It also requires returning the Merkle proofs along side the chunk so that it can be readily verified. Scheduling blocks for retrieval should largely remain the same, but additional request and response messages are required.
```

design/metadata-overhead.md
```diff
@@ -0,0 +1,212 @@
...
+
+#### Announcing over the DHT
+
+Also, datasets are now announced by their Merkle root instead of each individual block as was the case in the previous implementation. Announcing individual blocks is still supported, for example manifests are announced exactly the same as before, by their cid. Announcing individual blocks is also supported (but not required) and can be usefull in the case of bandwidth incentives.
```
markspanbroek, line 67:
```suggestion
Also, datasets are now announced by their Merkle root instead of each individual block as was the case in the previous implementation. Announcing individual blocks is still supported, for example manifests are announced exactly the same as before, by their cid. Announcing individual blocks is also supported (but not required) and can be useful in the case of bandwidth incentives.
```

markspanbroek, line 67: Would a Codex node that stores only the data for a single slot, still announce that data under the merkle root of the entire dataset?

design/metadata-overhead.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Block Stores and Local Repo
+
+All interaction with blocks/chunks sit behind the `BlockStore` abstraction, which currently only supports querying blocks by hash. It should be extended to allow querying by Merkle root and block index and/or range.
```
markspanbroek, line 71:
```suggestion
All interactions with blocks/chunks sit behind the `BlockStore` abstraction, which currently only supports querying blocks by hash. It should be extended to allow querying by Merkle root and block index and/or range.
```

design/metadata-overhead.md
dryajov: Good question - yes, so far I don't see a better way of doing so. There isn't a good way of announcing block ranges over a DHT, but even in that case, blocks in slots aren't continuous. Definitely not ideal, and something to investigate further.

dryajov:
> @benbierens I think that 2 elements per node is optimal when it comes to the size of the proofs. Think about the extreme case when you have N leaves and N elements per node. The size of the proof for each of the N leaves will be O(N).

This, the size of proofs increases with the arity of the tree.

> @dryajov I like the Idea. I'm not convinced yet that we need to store the entire tree. I think we could (should) store only a merkle proof next to a block and that should suffice for all the use cases in my opinion.
>
> It will take only 2 times more on the disk, but I think it would reduce the complexity of the flows and possibly increase the performance of disk operations (sequential reads only).

The problem with this approach is that, in the new flow, you only know the root of the tree and the index of the leaf, you don't know the actual leaf or block hash before hand, the merkle tree serves as a kind of an index, where you read the tree and fetch the leaf (block hash) and after you can read it from the repo.

PR #188, opened 2023-11-02
Add queue pausing and prevention of small availabilities from clearing the queue

emizzle:


design/sales.md
```diff
@@ -174,6 +174,7 @@ Slots in the queue should be sorted in the following order:
 2. Collateral required (ascending)
 3. Time before expiry (descending)
 4. Dataset size (ascending)
+5. Seen flag
```
AuHau, line 177: That should be first ordering criteria, so it works as you propose ;-)

AuHau (approved):
So, it feels like a workaround patch, and in my humble opinion, my proposal would better suit the functionality we are seeking. However, I agree that we don't have time for that big refactor now. So, let's go with this, and hopefully, we can revisit my original proposal later on. Possibly, we might identify some new data points from Testnet testing, which might affect the future implementation as well.

markspanbroek (approved):
Looks good to me, when the changes that Adam proposes are incorporated.

PR #189, opened 2023-11-06
initial Merkle tree doc draft

bkomuves:
Merkle tree variations to be discussed

### DIFF
--- a/design/Merkle.md
+++ b/design/Merkle.md
@@ -26,7 +26,7 @@ The hash function `H` can also have different types `S` of inputs. For example:
 - A naive Merkle tree implementation could for example accept only a power-of-two 
   sized sequence of `T`
 
-Notation: Let's denote a sequence of `T` by `[T]`.
+Notation: Let's denote a sequence of `T`-s by `[T]`.
 
 ### Merkle tree API
 
@@ -65,7 +65,7 @@ The compression function could be implemented in several ways:
 When implemented without enough care (like the above naive algorithm), there are several 
 possible attacks producing hash collisions or second preimages:
 
-1. The root of particular any layer is the same as the root of the input
+1. The root of any particular layer is the same as the root of the input
 2. The root of `[x_0,x_1,...,x_(2*k)]` (length is `n=2*k+1` is the same as the root of 
    `[x_0,x_1,...,x_(2*k),dummy]` (length is `n=2*k+2`)
 3. when using bytes as the input, already `deserialize` can have similar collision attacks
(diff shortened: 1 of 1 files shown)


design/Merkle.md
```diff
@@ -0,0 +1,226 @@
...
+**Solution 1.** Just prepend the data with the length `n` of the input sequence. Note that any
+cryptographic hash function needs an output size of at least 160 bits (and usually at least 
+256 bits), so we can always embed the length (surely less than `2^64`) into `T`. This solves
+both problems 1) and 2) (the height of the tree is a deterministic function of the length),
+and 4) too.
+However, a typical application of a Merkle tree is the case where the length of the input
+`n=2^d` is a power of two; in this case it looks a little bit "inelegant" to increase the size
+to `n=2^d+1`, though the overhead with above even-odd construction is only `log2(n)`.
+An advantage is that you can _prove_ the size of the input with a standard Merkle inclusion proof.
+Alternative version: append instead of prepend; then the indexing of the leaves does not change.
```
markspanbroek, line 101: Appending has another advantage; it makes it easier to deal with scenario's where the length isn't known beforehand, such as when streaming the data.

design/Merkle.md
```diff
@@ -0,0 +1,226 @@
...
+`512-64-1=447` bits of additional information; so for example
+
+    compress_k(x,y) := SHA256(k|x|y)
+
+works perfectly well with no overhead compared to `SHA256(x|y)`.
+
+**MiMC.** MiMC is another arithmetic construction, however in this
+case the starting point is a _block cipher_, that is, we start with
+a keyed permutation! Unfortunately MiMC-p/p is a (keyed) permutation 
+of `F`, which is not very useful for usl; however in Feistel mode we
```
markspanbroek, line 189:
```suggestion
of `F`, which is not very useful for us; however in Feistel mode we
```

markspanbroek (commented):
Thanks for taking the time to write this out @bkomuves :pray:

design/Merkle.md
```diff
@@ -0,0 +1,226 @@
...
+3. when using bytes as the input, already `deserialize` can have similar collision attacks
+4. The root of a singleton sequence is itself
+
+Traditional (linear) hash functions usually solve the analogous problems by clever padding.
+
+### Domain separation
+
+It's a good practice in general to ensure that different constructions using the same 
+underlying hash functions will never produce the same output. This is called "domain separation",
+and it's a bit similar to _multihash_; however instead of adding extra bits of information to a hash
```
dryajov, line 77: This is a bit of a misleading statement, `multihash` isn't a cryptographic construct at all, it is a meta-protocol for automatic negotiation in communication protocols.

bkomuves: What I meant (but it could be probably phrased better), is that since in multihash you have some extra information telling you _which_ hash you use, it can never happen that you confuse two differently constructed hashes. 

Domain separation achieves the same (probabilistically), by putting this information into the hash. So for example you cannot confuse sponges with different rates (it could only happen for the empty input anyway, but still).

dryajov: > What I meant (but it could be probably phrased better), is that since in multihash you have some extra information telling you _which_ hash you use, it can never happen that you confuse two differently constructed hashes.
> 
> Domain separation achieves the same (probabilistically), by putting this information into the hash. So for example you cannot confuse sponges with different rates (it could only happen for the empty input anyway, but still).

Right, but this are different things in the sense that, the domain separator is never part of the resulting hash, where `multihash`, exist specifically to identify the type of hash being transmitted or stored and it is always packed with the hash and can be interpreted later on. 

I mean, I understand the analogy you're trying to make, but I think it adds confusion because the two constructs serve completely different purposes. No biggie, but I had to stop and think what you meant in that section.

bkomuves: I changed the language to make it hopefully more clear that this is only a very loose analogy (i still think it's useful to mention it).

dryajov (approved):
LGTM, thanks for adding the chosen approach.

PR #190, opened 2024-05-07
Slot reservations

AuHau:


design/slot-reservations-simplified.md
```diff
@@ -0,0 +1,282 @@
...
+storage.
+
+#### Fill reward versus request collateral
+
+There is one caveat to the fill reward: if the fill reward is larger than the
+required collateral in an active request, an SP that is actively filling a slot
+will see a more profitable opportunity with a high fill reward (assuming SP's
+address was close to the source of the expanding window), and would be
+incentivized to abandon their existing duties and fill the slot in the new
+request.
```
AuHau, line 119: I doubt that this is an issue. I think Mark mentioned in one of our discussions that the fill reward is "secondary" here. The main goal for SPs is the request's reward... Also, the fill reward should be "small enough". Honestly, can't imagine that somebody would give out some large amount of money for it as the "main chunk" should go to reward.

Honestly, my question is should we set some lower bound of "minimum fill reward"? Should we support "zero fill reward"? I am bit thinking if we should not instead of leaving it up to the customer to set, to have some formula that would calculate it....

emizzle:
I've updated this proposal with the "4th proposal" as discussed in the recent meeting. This proposal is simplified to have: slot reservations, an expanding window mechanism, no fill reward, and no collateral.

design/slot-reservations.md
```diff
@@ -0,0 +1,225 @@
...
+            |                  |
+            v                  v
+          slot                slot
+         reserved            filled
+
+
+            ---------------- time ---------------->
+```
+
+There is an initial race for eligible SPs who are first to secure a reservation, then a
```
AuHau, line 45: > There is an initial race for eligible SPs who are first to secure a reservation

This race is not really related to our network as it is "settled" on blockchain. So IMHO not really an issue.

design/slot-reservations.md
```diff
@@ -0,0 +1,225 @@
...
+Adding a fill reward for SPs who ultimately fill the slot is not necessary
+because, like the SP rewards for providing proofs, fill rewards would be paid
+once the storage request successfully completes. This would mean that the fill
+reward is effectively the same as an increase in value of the storage request
+payout. Therefore, if a client is so inclined to provide a fill reward, they
+could instead increase the total reward of the storage request.
+
+In this simplified slot reservations proposal, there will not be reservation
+collateral and reward requirements until the behavior in a live environment can
+be observed and it is determined this are necessary mechanisms.
```
AuHau, line 105: Also we said that the "get payed for actual time hosting the slot" would also align with the "fill reward" even though most probably not in its full "force".

See: https://github.com/codex-storage/codex-contracts-eth/issues/102

design/slot-reservations.md
```diff
@@ -0,0 +1,225 @@
...
+number of eligible SP addresses increases, until eventually all SP addresses in
+the network are eligible.
+
+The expanding window mechanismstarts off with a random source address, defined
+as $hash(block hash, request id, slot index, reservationindex)$. The distance
+between each SP address and the source address can be defined as $XOR(A, A_0)$
+(kademlia distance). Once the allowed distance is greater than SP's distance,
+the SP is considered eligible to reserve a slot. The allowed distance for
+eligible addresses over time $t_i$ can be [defined as $2^{256} *
+F(t_i)$](https://hackmd.io/@bkomuves/BkDXRJ-fC). As this allowed distance value
```
AuHau, line 66:
```suggestion
eligible addresses over time $t_i$ can be defined as $2^{256} *
F(t_i)$([more information](https://hackmd.io/@bkomuves/BkDXRJ-fC)). As this allowed distance value
```

design/slot-reservations.md
```diff
@@ -0,0 +1,225 @@
...
+
+$XOR(A, A_0) < 2^{256} * F(t_i)$
+
+Because the source address for the expanding window is generated using the slot
+index and the reservation index, that means the source address for each
+reservation in each slot will be different.
+
+Furthermore, the client can set the rate of expansion by defining the [parameter
+$h$](https://hackmd.io/@bkomuves/BkDXRJ-fC#Parametrizing-the-speed-of-expansion)
+as the [dispersal parameter](https://github.com/codex-storage/codex-research/blob/ad41558900ff8be91811aa5de355148d8d78404f/design/marketplace.md#dispersal) of the request.
```
AuHau, line 78:
```suggestion
Furthermore, the client can set the rate of expansion by defining the parameter
$h$ ([more information](https://hackmd.io/@bkomuves/BkDXRJ-fC#Parametrizing-the-speed-of-expansion))
as the [dispersal parameter](https://github.com/codex-storage/codex-research/blob/ad41558900ff8be91811aa5de355148d8d78404f/design/marketplace.md#dispersal) of the request.
```

AuHau (commented):
Generally looks good to me! Just a few comments.

Also, I would suggest moving Balazs's write-up here as it is currently protected from non-logged-in access, and IMHO, we should centralize these documents...

P.S.: Can't approve my own PR, so somebody else will have to do that.

design/slot-reservations.md
```diff
@@ -0,0 +1,225 @@
...
+
+In this simplified slot reservations proposal, there will not be reservation
+collateral and reward requirements until the behavior in a live environment can
+be observed and it is determined this are necessary mechanisms.
+
+### Slot reservation attacks
+
+Name         | Attack description
+:------------|:--------------------------------------------------------------
+Clever SP    | SP drops reservation when a better opportunity presents itself
```
markspanbroek, line 111:
```suggestion
Clever SP    | SP drops slot when a better opportunity presents itself
```

markspanbroek (approved):
Very nice, thanks for taking the time to update this! I only have a few minor comments.

design/slot-reservations.md
emizzle: True, there is not an impact on the Codex network, but it describes chronologically how an SP would fill a slot once eligible, no?

design/slot-reservations.md
emizzle: Added additional comments to that effect.

design/slot-reservations.md
emizzle: I updated this section a bit, and removed any links on equations.

design/slot-reservations.md
emizzle: I updated this section a bit, and removed any links on equations.

emizzle:
> Also, I would suggest moving Balazs's write-up here as it is currently protected from non-logged-in access, and IMHO, we should centralize these documents...

Good point. @bkomuves, is it OK if we copy [your write up](https://hackmd.io/@bkomuves/BkDXRJ-fC) in here as part of the slot reservations design? 

If so, would you be able to provide the source to me, so that I don't have to redo all the equations manually? 🙏

design/slot-reservations.md
AuHau: Agree, but maybe I would only add a comment that first race is on blockchain, while second race is already on Codex network...

bkomuves:
@emizzle sure, feel free to copy the relevant parts (maybe edit down a little, for example you probably won't need the "how not to do it" section :) 

(you don't have access to the source? Not complete sure, but I think I can "invite" you to edit, DM me your hackmd username)

design/slot-reservations.md
emizzle: The second race involves downloading data from the network (likely only those with a reservation would download), but it is still a race amongst reserved SPs to fill the slot on chain. I actually think it's an important point to describe that there are races, as this was a question that someone had when we initially presented the idea at the offsite, and perhaps it may clarify how it works for other readers.

PR #191, opened 2024-05-29
Evaluation of Eigenlayer and EIGEN token

markspanbroek:
A review of the Eigenlayer and EIGEN token whitepapers, with some thoughts on
how this could be applied to Codex

### DIFF
--- a/evaluations/eigenlayer.md
+++ b/evaluations/eigenlayer.md
@@ -1,0 +1,112 @@
+Eigenlayer
+==========
+
+2024-05-29
+
+A review of the Eigenlayer and EIGEN token whitepapers, with some thoughts on
+how this could be applied to Codex.
+
+* [Eigenlayer whitepaper](https://docs.eigenlayer.xyz/assets/files/EigenLayer_WhitePaper-88c47923ca0319870c611decd6e562ad.pdf)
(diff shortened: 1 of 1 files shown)


emizzle (approved):
Very nice, mark!

leobago:
Nice evaluation @markspanbroek. I haven't read the whitepaper but looking at your writeup, a few questions came up.

If a validator restakes in another protocol, let's call it protocol A, and gets slashed by that protocol but not by Ethereum, the validator can still keep running on Ethereum because Eigenlayer cannot enforce slashing on Ethereum, right? I understand that when the validator exits, its stake will go directly to Eigenlayer because he was slashed in protocol A. However, the validator can decide to keep running as, from the Ethereum perspective, he hasn't done anything wrong. If this is the case, the validator can delay the payment of the stake as much as he wants, right? 

Like in the case of Codex, if an SP does restacking for his collateral, and then later on it misses a slot, we usually use part of its collateral to pay for the reconstruction by another SP. But in this case with Eigenlayer, the validator/SP can keep running on Ethereum, and the collateral will only be available once the validator decides to exit or gets slashed in Ethereum. This will be an issue. 

There might be more similar cases.

markspanbroek:
Good point @leobago, I updated the document with your insight that we cannot use restaked value for repair.

PR #192, opened 2024-06-13
Evaluation of L2 rollups

markspanbroek:
Quick and dirty overview of existing rollups and their suitability for hosting the Codex marketplace smart contracts.


evaluations/rollups.md
```diff
@@ -0,0 +1,219 @@
...
+rollup:
+1. Low gas costs; if gas is too costly then the business case of storage
+   providers disappears
+2. EVM compatibility; this shortens our time to market because we already have
+   Solidity contracts
+3. Support for BN254 elliptic curve precompiles (ecAdd, ecMul, ecPairing) for
+   the proof system
+4. High throughput; our current proof system that checks all proofs separately
+   on chain requires a large number of transactions per second
+5. Censorship resistant; an L2 operator should not have the power to exclude
```
dryajov, line 29: This is a nice to have, but not an absolute requirement given the current state of rollup ecosystems in general. It is well understood that right now, most of this projects are taking a pragmatic approach and so are we.

dryajov:
Great analysis, thanks for putting this together! 

I think reducing the number of proofs per node is paramount as a way forward with this approach. This is doable with node level proof aggregation, which would reduce the ~100 proofs per node per 24 hours, to a further 1 proof per node.

evaluations/rollups.md
markspanbroek: I agree; the state of the ecosystem doesn't give us much choice.

evaluations/rollups.md
```diff
@@ -0,0 +1,219 @@
...
+able to launch  a storage network without it. Reducing the number of proofs also
+ensures that the network can grow to a larger total size.
+
+When we look at reducing the transaction costs, then the best thing to focus on
+is getting rid of the need to post transactions on L1 in blobs. This is by far
+the most expensive part of running a rollup, and this is most likely also why
+Mantle is the only rollup in this overview that is cheap enough; it uses EigenDA
+instead of posting to L1. In this respect, it might also be interesting to look
+at Arbitrum AnyTrust, which has a similar design. We could also consider
+creating a fork of an existing rollup and use Codex as the DA layer. Some of the
```
leobago, line 214: Wouldn't this be a bit of a circular issue? Codex relies on the rollup to work, and the rollup relies on Codex to work.

markspanbroek: I think that you can do it like this (rough idea):
1. rollup creates block number x
2. rollup creates block number x+1, and includes a storage request for block x
3. rollup creates more blocks....
4. the storage request for block x is started, all it slots are filled
5. validators can now check block x...
6. if after 7 days no fraud proofs were detected, then block x is considered final

If you use want to use zk proofs of the block instead of fraud proofs, then you can consider block x final at step 4.

dryajov (approved):

dryajov:
I think the next step is to explore the more modular approaches that you mentioned in the writeup.

PR #193, opened 2024-07-18
Evaluation of side chains

markspanbroek:
A look at the economics of running the Codex marketplace on a side chain, either by using an existing one, or by creating a dedicated side-chain for Codex.


emizzle (approved):
Great analysis as always, Mark.

What about the economics of *starting* a side chain. My assumption is that there are two costly challenges: bringing validators online and maintaining bridges. 

Bootstrapping a side chain will require a minimum number of validators to participate in securing the network. If there is low participation at the start, then Codex would be forced to run validators to secure the network. The *profitability* (re: cost/revenue ratio) of that effort is highly dependent on the token price, which means running validators could potentially be a significant cost to Codex. Additionally, if profitability is low, it would disincentivise external validator participation.

Along the same lines, maintaining a bridge incurs a cost (of which I don't know) and there is an [increased security risk](https://medium.com/immunefi/common-cross-chain-bridge-vulnerabilities-d8c161ffaf8f) when running a bridge.

Additionally, there may be additional marketing and community management costs to start and maintain the side chain.

leobago:
Nice analysis @markspanbroek .

Quick question: How many validators per node are you assuming in your analysis?

I believe Gnosis can run many validators in one node, and therefore, the hardware cost can decrease significantly.

Typo:  more transactions than fit in the blocks => that fit in....

markspanbroek:
> Quick question: How many validators per node are you assuming in your analysis?
> 
> I believe Gnosis can run many validators in one node, and therefore, the hardware cost can decrease significantly.

That is a good question @leobago, I assumed one validator per machine. I never thought about running multiple validators on a single machine because it defeats the purpose of having many validators: decentralization.

> Typo: more transactions than fit in the blocks => that fit in....

I did mean "than" as in "more than". As soon as the amount of transactions in the network exceed the capacity of the blocks, then there is scarcity and the price will go up.

